PR #5, opened 2017-01-14
[IMP] Add env var expansion for repo.yml

lasley:
* Expand environment variables in repo.yml before use

My use case being a dynamic odoo version, which interestingly enough could also work for module branches too.

### DIFF
--- a/build.d/20-repos-aggregate
+++ b/build.d/20-repos-aggregate
@@ -11,8 +11,9 @@ conf=/opt/odoo/custom/src/repos.yaml
 if [ -f $conf ]; then
     log INFO Aggregating repositories from $conf
     cd $(dirname $conf)
-    echo "$(cat $conf | envsubst)" > $conf
-    gitaggregate -c $conf
+    envsubst < $conf > /tmp/repos.yaml
+    gitaggregate -c /tmp/repos.yaml
+    rm /tmp/repos.yaml
 else
     log ERROR Cannot aggregate repositories: $conf not found
     exit 1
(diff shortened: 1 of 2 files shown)


build.d/20-repos-aggregate
```diff
@@ -11,6 +11,7 @@ conf=/opt/odoo/custom/src/repos.yaml
 if [ -f $conf ]; then
     log INFO Aggregating repositories from $conf
     cd $(dirname $conf)
+    echo $(eval echo $(cat $conf)) > $conf
```
Yajo: Better use envsubst, like with conf files.

lasley:
> Don't you think PR for gitaggregator would be better?

Env vars seem a bit more on the side of the system configuration vs. a CLI tool IMO, but I could be wrong. TBH this is the first I've heard of the tool, so I'm not sure of a use case outside of this one. If you think I should submit that way though, I don't have any problems with it.

TIL with envsubst - not sure how I've gone this long in life without needing it! Updated.

Yajo:
I added a little fix, but there's still a problem: this would break [the scaffolding's setup-devel step](https://github.com/Tecnativa/docker-odoo-base#development).

We could either support this also in [tecnativa/git-aggregator](https://hub.docker.com/r/tecnativa/git-aggregator) or you could propose this PR to [upstream](https://github.com/acsone/git-aggregator).

lasley:
Hrmmm alright so out of the choice of git-aggregator Docker container vs. library, library is probably best. Seems like the feature would get buried. PR in https://github.com/acsone/git-aggregator/pull/8

Yajo:
Nice, I merged acsone/git-aggregator#8 in our fork's master branch, so you have it in https://hub.docker.com/r/tecnativa/odoo-base/~/settings/automated-builds/

lasley:
Hooray thanks @Yajo! This feature should help me quite a bit 😄

Issue #6, opened 2017-01-18
Restore tests


Yajo:
This repo needs CI per pull request.

Current testing in docker hub fails for unknown reasons ([seems like the unittest script exits with a non-zero code, although tests pass](https://hub.docker.com/r/tecnativa/odoo-base/builds/b3rxqzxyeohuooe6zdve27f/)).

I removed tests in https://github.com/Tecnativa/docker-odoo-base/commit/b1308d1378c1348d120a2a7e32e58c1620256d63, I have to figure out how to apply a real CI here and how to make them work, but for now I need this image to be built.

lasley:
Should we merge this and #14?

Yajo:
These are issues, not PR, what do you mind by merging?

lasley:
> what do you mind by merging?
Well there are two tickets for the same thing. Was just wondering if it made sense to leave the two open or just work under one? The management overhead of two tickets just seems meh 😉

PR #15, opened 2017-02-18
Add postgresql repository

carlosdauden:
Add postgresql repository to get the latest version of postgresql-client

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,9 +70,14 @@ RUN apt-get update \
         libxml2 libxslt1.1 libjpeg62-turbo zlib1g libfreetype6 liblcms2-2 \
         libopenjpeg5 libtiff5 tk tcl libpq5 libldap-2.4-2 libsasl2-2 \
         # This image's facilities
-        bzip2 curl gettext-base git nano npm openssh-client \
-        postgresql-client telnet \
+        bzip2 curl gettext-base git nano npm openssh-client telnet \
     && curl https://bootstrap.pypa.io/get-pip.py | python /dev/stdin --no-cache-dir \
+
+    # Special case to get latest PostgreSQL client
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I think the idea is to base this on Alpine as soon as we get rid of the nasty bug we have, so I'm not sure this is the best option for long term.

blaggacao:
For alpine it goes arround this

```
RUN echo "@edge http://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories \
    && apk --no-cache add "postgresql@edge<9.7" "postgresql-contrib@edge<9.7"
```

lasley:
Update on the stack overflow - I'm still able to replicate it using our email templates, even with our patch. We may be a ways off from Alpine 😦 . At this point, I'm 99% that lxml is the culprit - it's the only lib involved in everything that fails. But then again, Odoo uses lxml for basically everything, so it makes sense it's in all the stacks.

Dockerfile
```diff
@@ -61,6 +61,8 @@ ENV OPENERP_SERVER=/opt/odoo/auto/odoo.conf \
 
 # Other requirements and recommendations to run Odoo
 # See https://github.com/$ODOO_SOURCE/blob/$ODOO_VERSION/debian/control
+RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ xenial-pgdg main' >> /etc/apt/sources.list.d/postgresql.list
```
Yajo: You are adding the ubuntu version of the repos, please use Debian's if available.

Yajo:
Hi all, I added #16 to track all Alpine stuff, so we have some meeting point about it and have no more need to keep on talking about it on every issue.

About the PR itself, just one requirement.

BTW, since we still did not fix #6, did you try building this Dockerfile locally and did it work, @carlosdauden?

PR #25, opened 2017-04-28
[FIX] YAML to YML

lasley:
* Fix file extension and references to YML

While technically `.yaml` is a valid file extension for YAML, the commonly used one is `.yml` - especially in Docker realm. This proposes that change.

I also made acsone/git-aggregator#14, although technically that just allows it into auto-complete in the shell. You could already use any file format AFAIK.

I will submit another change to your scaffold too once this is accepted.

### DIFF
--- a/lib/odoobaselib.py
+++ b/lib/odoobaselib.py
@@ -6,11 +6,16 @@ import yaml
 
 # Constants needed in scripts
 SRC_DIR = "/opt/odoo/custom/src"
-ADDONS_YAML = SRC_DIR + "/addons.yaml"
+ADDONS_YAML = '%s/addons' % SRC_DIR
 ADDONS_DIR = "/opt/odoo/auto/addons"
 CLEAN = os.environ.get("CLEAN") == "true"
 LINK = os.environ.get("LINK") == "true"
 LOG_LEVELS = ("DEBUG", "INFO", "WARNING", "ERROR")
+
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Well, this supposes that we have to change all our Docker projects (they are a lot right now) to rename the files, without no real benefit. Maybe it's not the most commonly used extension, but if it's supported, why not?

I wanna keep the current extension unless you have stronger reasons.

lasley:
Kind of dirty, but what about 6a32bf1

pedrobaeza:
OK, that sounds better. @Yajo will give next Tuesday the final judgment

lib/odoobaselib.py
```diff
@@ -6,11 +6,16 @@
 
 # Constants needed in scripts
 SRC_DIR = "/opt/odoo/custom/src"
-ADDONS_YAML = SRC_DIR + "/addons.yaml"
 ADDONS_DIR = "/opt/odoo/auto/addons"
+ADDONS_YAML = '%s/addons' % SRC_DIR
 CLEAN = os.environ.get("CLEAN") == "true"
 LINK = os.environ.get("LINK") == "true"
 
+if os.path.isdir('%s.yml' % ADDONS_YAML):
```
Yajo: `isdir`?

lasley:
Whoops, seems I forgot about this! Bug fixed & `yaml` is now the default again.

PR #27, opened 2017-06-10
[IMP] Create an addons wildcard

lasley:
This adds a wildcard for the symlinking of modules directories in the `addons.yaml` file, which works out really great for dev instances IMO (or for lazy people, like me).

I tested manually, but didn't really see an obvious way to test via CI. Am I missing something, perhaps?

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,13 @@ server-tools: # Here we repeat server-tools, but no problem because it's a
     - html_text
 ```
 
+You can add all modules in a repo by using a `*`:
+
+```yaml
+website:
+    - "*"
+```
+
 Important notes:
 
 - Do not add repos for the required [`odoo`][] and [`private`][] directories;
@@ -185,6 +192,8 @@ Important notes:
   removed by default, to keep the resulting image thin).
 
 - If you list 2 addons with the same name, you'll get a build error.
+
+- If you use the wildcard (`*`), it must be encapsulated in quotes.
 
 ##### `/opt/odoo/custom/src/requirements.txt`
 
--- a/entrypoint.d/40-addons-link
+++ b/entrypoint.d/40-addons-link
@@ -6,10 +6,6 @@ import sys
 
 import odoobaselib
 
-if not odoobaselib.LINK:
-    logging.warning("Not linking addons")
-    sys.exit()
-
 logging.info(
     "Linking all addons from %s in %s",
     odoobaselib.ADDONS_YAML,
--- a/entrypoint.d/40-addons-link
+++ b/entrypoint.d/40-addons-link
@@ -30,5 +30,6 @@ for repo, addons in odoobaselib.addons_config().items():
             continue
         src = os.path.join(odoobaselib.SRC_DIR, repo, addon)
         dst = os.path.join(odoobaselib.ADDONS_DIR, addon)
-        logging.debug("Linking %s in %s", src, dst)
-        os.symlink(src, dst)
+        if not os.path.exists(dst):
+            logging.debug("Linking %s in %s", src, dst)
+            os.symlink(src, dst)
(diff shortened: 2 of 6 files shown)


pedrobaeza:
This option was discarded by @Yajo because you can't link before having downloaded the repos and mounted the filespace, but I don't know if you have found a solution for that.

lasley:
This step is injected during the symlink phase, before the unused modules are discarded. Without this, there are a bunch of messages about module deletions & with this those messages are no longer present. I admittedly haven't played with the generated box quite yet, because now I'm working on #28, but it looked good when I inspected the file system after build.

Yajo:
Thanks @lasley! But please check https://github.com/Tecnativa/docker-odoo-base/issues/26 out. It is a requirement for us to support devel, test and prod environments (it's one of this project's main strengths).

In devel, build speed is most important; that's why we don't actually copy the code in the image (after all it will be mounted). Since no code is copied, there's no way to know what `*` expands to, and `build.d` is executed at build time.

If you really need this, the only option would be moving the symlinking process to `entrypoint.d`, which gets executed at run time and always has the code, be it either from the volumes or from the frozen production image.

Definitely we need some CI, you're more expert on that than us, do you think you could PR us with some of that?

lasley:
@Yajo - Hmmm ok I see the issue. I haven't messed too much with the different scaffolds yet, because we are not able to build directly in our infrastructure. We have to work from  pre-built images, so I've been manually running things from the Dockerfile.

I do need this unfortunately. Are you good with moving the entire addons linking process into the entrypoint in order to keep from splitting it up? It looks like we already have an isolation between that and the cleanup process, so I don't think it would hurt anything to not have the symlinks until start. Symlinks are pretty light in terms of overhead too, so I don't think this would stall boot too much.

README.md
```diff
@@ -162,6 +162,13 @@ server-tools: # Here we repeat server-tools, but no problem because it's a
     - html_text
 ```
 
+You can add all modules in a repo by using a `*`:
+
+```yaml
+website:
+    - "*"
```
Yajo: Hmm I feel the right design would be to use it as a single string. Possible options:

```
website: "*"
web: all
server-tools: true
```

From the above, please choose only one (the one you like most, I don' really care 😋). Setting it as an array item as you purpose makes not much sense IMHO.

entrypoint.d/40-addons-link
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import logging
+import os
+import sys
+
+import odoobaselib
+
+if not odoobaselib.LINK:
```
Yajo: `LINK` is currently a build arg, so it will never be there at run time. You need to change that in the Dockerfile and possibly in the docs.

However, thinking about it twice, if we move the symlinking process to the entrypoint, then it makes not much sense to add the feature of *not* linking (this made sense when we had the problem said before, but not with this patch...). We can just remove this (remember: here, in Dockerfile and in README if it was there).

I love removing things! 😊

README.md
lasley: Works for me - I'll go with `website: "*"` so that the code works without changes, due to strings being treated as arrays.

Yajo: Another option I'm thinking of right now (sorry for annoying): fully support globs. This way you can:

```yaml
event:
  - event_*
website:
  - *
```

I think I like this one most!

lasley: Hahaha ok so glob support was my first thought, which ended up being pretty damned difficult. In hindsight, I probably should have just made it a regex & replaced the `*` with `.*?` instead. Will give that a shot

lasley: Ohhhh ok I remember why I didn't implement this - it will require a redesign of the way that the logic works in some of the loops. I admittedly didn't want to get into that yet because I don't need it yet.

This is also why I left it at the current syntax though, which will allow for an easier feature addition.

Are you opposed to just adding this to Roadmap for in incremental improvement & leaving the current syntax as is?

Yajo: OK, it's good as it is right now then 😉

lasley:
Thanks for the consideration guys, even if it's not something you're going to use 😄

PR #29, opened 2017-06-12
[IMP] Add Basic CI

lasley:
This is a WIP adding basic CI to this repo. TBD whether it actually works in this context- I just took it directly from our Odoo Docker image for the most part.

May want to go ahead and turn Travis on for this repo so we can see results in the PR, but this is the [project on our end](https://travis-ci.org/LasLabs/docker-odoo-base)

PS - no judgement allowed about the fact that our DQT (Docker Quality Tools) doesn't use compose files. It was a stupid design choice that we will fix eventually.


Yajo:
Last time I looked into this, I loved [Docker Cloud's design on testing](https://docs.docker.com/docker-cloud/builds/automated-testing/). Do you think we can imitate that here, but in a different fashion?

I was thinking on having a `tests/` top folder, full of `.yaml` files. Each one of those would be a docker compose file that includes a container called `sut`, which will be ran and should exit with error if our test case failed.

So all Travis should do is:

1. Install latest `docker-compose`.
2. Something similar to `for file in tests/*; do ( docker-compose -f $file run --rm sut || exit 1 ) && docker-compose -f $file down -v --rmi local; done`.

The reason for *not* using a `/docker-compose.test.yaml` file as Docker Cloud would do is that tests would be triggered automatically on Docker Hub builds, and those are slow enough right now. I prefer to test in Travis only.

What do you think?

PS: Linking to #6.

lasley:
Hrmmmm this would require a pretty hefty redesign of the way that our pre-existing tools work, which are implemented in quite a lot of other images. I agree this isn't necessarily the best design around, but it's fairly tested in other images, includes a lint, and the ability to upload tags to DockerHub.

I'm looking at redesigning the tools at some point, but I think that it would be worth implementing as-is once I get these green - particularly with the requested PR process in #36.

pedrobaeza:
OK, go ahead with this then!

lasley:
Alright this is 🍏  - https://travis-ci.org/LasLabs/docker-odoo-base

lasley:
Basic tests are included here:
* [One that tests for the Odoo configuration's existance](https://github.com/LasLabs/docker-odoo-base/blob/8808284aa82b9dd0e560499a63f5c4b95529421a/tests/test_for_file)
* [One that tests Odoo web response](https://github.com/LasLabs/docker-odoo-base/blob/8808284aa82b9dd0e560499a63f5c4b95529421a/tests/test_curl)

.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+
+services:
+  - docker
+
+env:
+  global:
+    - PORTS="8069:8069"
+    - LINKS="db:db"
+    - DAEMONIZE="1"
+    - ODOO_VERSION="10.0"
```
lasley: @Yajo - I'm not sure of your Dockerhub deploy strategy, but this can handle that too. You'll need to do that if you want though - Take a look [at the sample Travis](https://github.com/LasLabs/docker-quality-tools/blob/master/sample_files/.travis.yml#L9). Variables to note:
* DOCKER_REPO
* DOCKER_USER
* DOCKER_TAG
* $DOCKER_PASS_TOKEN
* HUB

[An example of it in use](https://github.com/LasLabs/docker-alpine-jira/blob/master/.travis.yml#L20)

lasley: [Example of pushing multiple tags](https://github.com/LasLabs/docker-odoo/blob/master/.travis.yml#L21)

lasley:
This is ready for review

.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+
+  matrix:
+    - TESTS="1" ODOO_VERSION="10.0" DOCKER_CONTEXT="./tests/odoo10"
+    - TESTS="1" ODOO_VERSION="9.0" DOCKER_CONTEXT="./tests/odoo9"
+    - LINT_CHECK="1"
+
+before_install:
+  - docker build -t "tecnativa/odoo-base:${ODOO_VERSION}" --build-arg "ODOO_VERSION=${ODOO_VERSION}" ./
+  - docker pull kiasaki/alpine-postgres
```
Yajo: This is not the postgres version we support, it should be the official one `postgres:9.6-alpine`

.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+  - docker run -d -it -e POSTGRES_USER="odoo" -e POSTGRES_PASSWORD="odoopassword" --name="db" postgres:9.6-alpine
```
Yajo: Can't we use docker-compose instead of just docker? So we can simulate the scaffolding more closely.

tests/odoo9/Dockerfile
```diff
@@ -2,3 +2,14 @@ FROM tecnativa/odoo-base:9.0
...
+# Metadata
+ARG VCS_REF
+ARG BUILD_DATE
+ARG VERSION
+LABEL org.label-schema.schema-version="$VERSION" \
+      org.label-schema.vendor=Tecnativa \
+      org.label-schema.license=Apache-2.0 \
+      org.label-schema.build-date="$BUILD_DATE" \
+      org.label-schema.vcs-ref="$VCS_REF" \
+      org.label-schema.vcs-url="https://github.com/Tecnativa/docker-odoo-base"
```
Yajo: We don't need all of this.

tests/odoo10/Dockerfile
```diff
@@ -2,3 +2,14 @@ FROM tecnativa/odoo-base:10.0
...
+# Metadata
+ARG VCS_REF
+ARG BUILD_DATE
+ARG VERSION
+LABEL org.label-schema.schema-version="$VERSION" \
+      org.label-schema.vendor=Tecnativa \
+      org.label-schema.license=Apache-2.0 \
+      org.label-schema.build-date="$BUILD_DATE" \
+      org.label-schema.vcs-ref="$VCS_REF" \
+      org.label-schema.vcs-url="https://github.com/Tecnativa/docker-odoo-base"
```
Yajo: None of this is needed

lib/odoobaselib/__init__.py
```diff
@@ -38,6 +38,8 @@
 def addons_config():
     """Load configurations from ``ADDONS_YAML`` into a dict."""
     config = dict()
+    if not os.path.isfile(ADDONS_YAML):
+        return config
```
Yajo: [EAFP](https://docs.python.org/3/glossary.html#term-eafp). Use a try except block please 😊

.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+env:
+  global:
+    - PORTS="8069:8069"
+    - LINKS="db:db"
+    - DAEMONIZE="1"
+    - ODOO_VERSION="10.0"
+
+  matrix:
+    - TESTS="1" ODOO_VERSION="10.0" DOCKER_CONTEXT="./tests/odoo10"
+    - TESTS="1" ODOO_VERSION="9.0" DOCKER_CONTEXT="./tests/odoo9"
```
Yajo: What's this `TESTS` variable supposed to do?

lasley: Activates the tests, which just runs through the binaries in the `tests` folder. There is also `LINT_CHECK` which does the obvious, and `HUB` which pushes to DockerHub

.travis.yml
lasley: This would go back into changing our build tools, which TBH I don't really have time for at the moment. I feel that it would be best to merge as is, then polish things like this as we go forward. I agree this isn't the best design, but it works and it's better than what you have now.

tests/odoo9/Dockerfile
lasley: It's part of the build scripts, and is required. I never considered a use case where a child Dockerfile would be used, so the fact that the script fills these in never became an issue until now. It's another build tools change that I don't quite have time for TBH - this doesn't necessarily hurt anything & we most certainly use this aspect in our other builds.

Yajo:
I enabled travis and resent your last commit, just to trigger it

Issue #30, opened 2017-06-12
SSH Keys for Private Repos


lasley:
I am unable to get private repo SSH keys to work while building a scaffold. I've tried a lot of things, but the below is the most simple that definitely should have worked (it's in a `build.d` file):
```bash
#!/bin/bash
mkdir -p ~/.ssh
echo "RSA PRIVATE KEY" > ~/.ssh/id_rsa
echo "Host repo.laslabs.com
  IdentityFile ~/.ssh/id_rsa" > ~/.ssh/config
chmod 700 ~/.ssh/id_rsa
```

My guess here is that we aren't operating within a true shell, so the SSH config & key simply isn't being looked at. 

I've also tried to add this logic in my scaffold Dockerfile, but that's a bust because it's run after the repos are gathered.

Are you guys using any private repos at this point?

pedrobaeza:
Create a file with +x in odoo/custom/build/ with this content:
```
set -e
private='
-----BEGIN RSA PRIVATE KEY-----
...
-----END RSA PRIVATE KEY-----
'
public='
ssh-rsa ...== root@odoo
'

homes="/root /opt/odoo"
for dir in $homes; do
    dir=$dir/.ssh
    mkdir -p $dir
    echo "$private" > $dir/id_rsa
    echo "$public" > $dir/id_rsa.pub
    ssh-keyscan <git_host> >>$dir/known_hosts
    chmod -R u=rwX,go= $dir
    log INFO Added deployment keys to $dir
done
```
replace <git_host> with the name of your GIT private repos host and put the private and public keys. 

Then, you'll be able to add to repos.yaml and addons.yaml this private repo and addons.

Yajo:
Maybe it would be good to bundle support for this, given it is a common use: have a `id_rsa` and `id_rsa.pub` files/args, and a `GIT_HOSTS` build arg and add this bunch of code in a bundled build script...

lasley:
Thanks @pedrobaeza - exactly what I needed.

@Yajo - Yes, I am planning on submitting a PR to bundle support for this - your plan sounds solid too. This is totally common, and wasn't obvious enough IMO. I admittedly spent an hour or two before I gave up and submitted a ticket - I can only imagine how long that would have been for someone less experienced, or without a pre-existing relationship with the module maintainers.

lasley:
Reporting back - Pedro's script seems to have the same issue as mine - the private key isn't actually being used for the gitaggregate clones. 

[...]

PR #32, opened 2017-06-15
[IMP] Add SSH key logic

lasley:
This is my WIP proposal for #30


lasley:
Alright I fixed a few bugs & this looks to work now. I'd say ready for review.

Note on my design decision here: I do not feel that obtaining the SSH fingerprint on the fly is secure, thus requiring the inclusion of the `known_hosts` file. I also didn't feel that environment variables were the solution here due to the need to possibly support multiple keys and hosts.

pedrobaeza:
Why do you want to put something in root user?

And why fetching fingerprint is not secure?

lasley:
The root user seems to be the one doing the repo clones, but really I just took that from the fact that you did the same thing in the SSH script: `homes="/root /opt/odoo"`

The fingerprint is only good if it's known & hasn't changed. A MitM situation during the docker build cannot be detected if the key is being gathered each time it's used, and we might as well just add `StrictHostKeyChecking` to the config files in that case.

pedrobaeza:
About root, you're right. Then the question should be reversed: why adding odoo user?

About the known hosts, I personally see that for GIT clone operations as something very annoying, so if there's a way to disable that question, I prefer that path.

lasley:
> Then the question should be reversed: why adding odoo user?

I dunno TBH - you did that in the script as well, so I just copied it. Figured you knew something I didn't. Sounds like it's superfluous then?

> About the known hosts, I personally see that for GIT clone operations as something very annoying, so if there's a way to disable that question, I prefer that path.

This should be configured in the SSH config, like so:

```
Host repo.example.com
  StrictHostKeyChecking no
```

While I agree it's annoying, we require the security on our builds - transmission security and authenticity is important in the healthcare industry. We have to avoid the scenario where our repo is MitM'd and malicious code is injected into our customer instances.

pedrobaeza:
Yeah, right, I copied/pasted the script blindly without seeing that odoo user doesn't require the same keys if they are cloned with root.

About the known hosts, I think we can provide both ways (strict know_hosts file or fetching from domain) for satisfying both.

build.d/10-ssh-key
```diff
@@ -0,0 +1,33 @@
...
+
+
+if not os.path.isfile(CONFIG_PATH):
+    logging.info('No SSH configuration found')
+    exit(0)
+
+
+HOMES = {
+    'root': '/root',
+    'odoo': '/opt/odoo',
```
Yajo: No need for these, just `os.path.expanduser("~root")` as needed.

OTOH, as said, no need for adding deploy keys for odoo user.

build.d/10-ssh-key
```diff
@@ -0,0 +1,33 @@
...
+}
+
+
+def do_command(command_list):
+    proc = subprocess.Popen(command_list)
+    proc.communicate()
```
Yajo: This method fits better in `odoobaselib`

README.md
```diff
@@ -90,6 +91,24 @@ Executables here will run just before those in `/opt/odoo/common/build.d`.
 Files here will be environment-variable-expanded and concatenated in
 `/opt/odoo/auto/odoo.conf` at build time.
 
+#### `/opt/odoo/custom/ssh.d`
```
Yajo: The `.d` suffix in this project normally indicates that things inside it will be looped, either by `run-parts` or concatenanting all of them, or whatever.

This case is different, so just call it `ssh`.

README.md
```diff
@@ -90,6 +91,24 @@ Executables here will run just before those in `/opt/odoo/common/build.d`.
...
+
+Files here will be used to create the `root` and `odoo` users' SSH directories.
+
+It should follow the same structure as a standard `~/.ssh` directory, including
+`config` and `known_hosts` files.
```
Yajo: Since we concluded we don't need this for `odoo` user, I'm thinking on a different approach, just add this line to the Dockerfile:

```
RUN ln -s /opt/odoo/custom/ssh ~root/.ssh
```

And we're done!

Possibly this comment invalidates some of my comments below, but I'll keep them just in case. 😅

README.md
```diff
@@ -90,6 +91,24 @@ Executables here will run just before those in `/opt/odoo/common/build.d`.
...
+
+The `config` file should contain `IdentityFile` keys to represent the private
+key that should be used for that host. The key will be located in  `~/.ssh/`.
```
Yajo: You could add some docs links on how to use the ssh_config file.

Also some FAQs to document answer to @pedrobaeza's concerns and to document why it's more secure to hardcode the git server identity, based on your comments too.

PR #33, opened 2017-06-16
[IMP] Add abstract dependencies installation

lasley:
I think you guys might like this one. I haven't really tested much yet

* Add dependencies installation for apt, npm, gem & pip from files in custom/dependencies folder
* Support old requirements path

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,12 +55,17 @@ This is its structure:
         build.d/
         conf.d/
         ssh/
+        dependencies/
+            apt.txt
+            apt_build.txt
+            gem.txt
+            npm.txt
+            pip.txt
         src/
             private/
             odoo/
             addons.yaml
             repos.yaml
-            requirements.txt
     common/
         entrypoint.sh
         build.sh
@@ -229,10 +234,12 @@ Important notes:
 
 - If you use the wildcard (`*`), it must be encapsulated in quotes.
 
-##### `/opt/odoo/custom/src/requirements.txt`
+##### `/opt/odoo/custom/dependencies/pip.txt`
 
 A normal [pip `requirements.txt`][] file, to install dependencies for your
 addons when building the subimage.
+
+###
 
 ### `/opt/odoo/common`: The useful one
 
@@ -808,6 +815,11 @@ Dockerfile similar to this one:
 # Hash-pinned version of tecnativa/odoo-base:10.0
 FROM tecnativa/odoo-base@sha256:fba69478f9b0616561aa3aba4d18e4bcc2f728c9568057946c98d5d3817699e1
 ```
+
+### What is `/opt/odoo/custom/src/requirements.txt`?
+
+This is the deprecated file path for the PIP requirements. Use
+`/opt/odoo/custom/dependencies/pip.txt` instead.
 
 ### How can I help?
 
--- a/lib/installer.py
+++ b/lib/installer.py
@@ -4,14 +4,24 @@ import logging
 import os
 import subprocess
 
+from odoobaselib import do_command
+
 
 class Installer(object):
     """ Install all sorts of things. """
+
+    _HAS_RUN = {}
 
     def __init__(self, handler_name, file_path):
         self.handler_name = handler_name
         self.file_path = file_path
         self.requirements = self._parse_requirements()
+
+    def has_run(self, install_type, update=True):
+        has_run = self._HAS_RUN.get(install_type, False)
+        if update:
+            self._HAS_RUN[install_type] = True
+        return has_run
 
     def install(self):
         install_handler = '_install_%s' % self.handler_name
@@ -21,10 +31,23 @@ class Installer(object):
         remove_handler = '_remove_%s' % self.handler_name
         getattr(self, remove_handler)()
 
-    def _install_apt(self):
+    def cleanup(self):
+        remove_handler = '_cleanup_%s' % self.handler_name
+        getattr(self, remove_handler)()
+
+    def _cleanup_apt(self):
         self._run_command(
-            'apt-get update',
+            'apt-get -y autoremove',
         )
+        self._run_command(
+            'rm -Rf /var/lib/apt/lists/*',
+        )
+
+    def _install_apt(self):
+        if not self.has_run('apt'):
+            self._run_command(
+                'apt-get update',
+            )
         self._install_requirements([
             'apt-get',
             '-o', 'Dpkg::Options::="--force-confdef"',
@@ -34,12 +57,6 @@ class Installer(object):
             split=False,
             shell=True,
         )
-        self._run_command(
-            'apt-get -y autoremove',
-        )
-        self._run_command(
-            'rm -Rf /var/lib/apt/lists/*',
-        )
 
     def _remove_apt(self):
         for requirement in self.requirements:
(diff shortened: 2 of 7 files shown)


build.d/80-dependencies
```diff
@@ -0,0 +1,41 @@
...
+    'pip': 'pip install',
+    'npm': 'npm install -g',
+    'gem': 'gem install',
+}
+
+
+def do_install(command, requirements_file):
+    with open(requirements_file, 'r') as fh:
+        requirements = fh.read().strip()
+    command += ' %s' % requirements
```
Yajo: Yikes, this loop is interesting, but I feel each install method has its things.

For instance,

- `apt` needs an `apt-get update` before installing the build deps, and needs `apt-get -y autoremove && rm -Rf /var/lib/apt/lists/*` after installing run deps.
- `pip` can better use its `--requirements` flag instead of this read-and-append hack. (maybe not required, see next point).
- the rest of methods would benefit of removing lines from its requirements file that start by `#`, so we allow comments there (for pip this would mean the same as the point above).

`rm /tmp/*` would be great too at the end of everything.

build.d/80-dependencies
```diff
@@ -0,0 +1,41 @@
...
+    proc = subprocess.Popen(command.split(' '))
```
Yajo: if you call split without arguments, in case there are multiple spaces it will split correctly, and strip before:

```python
>>> "   asdfasdf  a sfdasdf    ".split(" ")
['', '', '', 'asdfasdf', '', 'a', 'sfdasdf', '', '', '', '']
>>> "   asdfasdf  a sfdasdf    ".split()
['asdfasdf', 'a', 'sfdasdf']
```

build.d/80-dependencies
```diff
@@ -0,0 +1,41 @@
...
+for name, command in REQ_TYPES.items():
+    req_file = os.path.join(
+        DIR_CUSTOM, 'requirements', '%s.txt' % name,
+    )
+    if os.path.isfile(req_file):
+        do_install(command, req_file)
+        if name == 'pip':
+            found_pip = True
+
+if not found_pip:
```
Yajo: You can drop this line and stick with the `if` below.

build.d/80-dependencies
```diff
@@ -0,0 +1,41 @@
...
+    req_file = '/opt/odoo/custom/src/requirements.txt'
+    if os.path.isfile(req_file):
```
Yajo: Add a deprecation comment here.

build.d/80-dependencies
lasley: But I wanted to be lazy! New code will be better 🚀

README.md
```diff
@@ -200,6 +205,8 @@ Important notes:
 A normal [pip `requirements.txt`][] file, to install dependencies for your
 addons when building the subimage.
 
+###
```
Yajo: This one is still empty.

lib/installer.py
```diff
@@ -0,0 +1,98 @@
...
+        install_handler = '_install_%s' % self.handler_name
+        getattr(self, install_handler)()
+
+    def remove(self):
+        remove_handler = '_remove_%s' % self.handler_name
+        getattr(self, remove_handler)()
+
+    def _install_apt(self):
+        self._run_command(
+            'apt-get update',
```
Yajo: I'd like to avoid running this twice if we have both apt and apt_build. Maybe something like `if not self.apt_updated: ...`?

lib/installer.py
```diff
@@ -0,0 +1,98 @@
...
+            '-y', '--no-install-recommends', 'install',
+        ],
+            split=False,
+            shell=True,
+        )
+        self._run_command(
+            'apt-get -y autoremove',
+        )
+        self._run_command(
+            'rm -Rf /var/lib/apt/lists/*',
```
Yajo: In the same line as above comment, if we have both apt and apt_build, this should go in the last step only, to avoid updating twice.

lib/installer.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+
+    def _remove_apt(self):
+        for requirement in self.requirements:
+            self._run_command(
+                'apt-get purge -y %s' % requirement,
```
Yajo: Add also an `apt-get autoremove -y` after the purge. It will remove all dangling dependencies too.

lib/installer.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _install_pip(self):
+        self._install_requirements(
+            'pip --no-cache-dir install',
+        )
+
+    def _install_requirements(self, command, split=True, shell=False):
+        if split:
+            command = command.split()
+        for requirement in self.requirements:
```
Yajo: I think all of the supported package managers allow passing all dependencies in a single command, can't we drop this loop?

lib/installer.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _remove_apt(self):
+        for requirement in self.requirements:
+            self._run_command(
+                'apt-get purge -y %s' % requirement,
+            )
+
+    def _install_gem(self):
+        self._install_requirements(
+            'gem install --no-rdoc --no-ri --no-update-sources',
```
Yajo: Add [gem cleanup](https://github.com/Tecnativa/docker-odoo-base/blob/master/Dockerfile#L97) after this.

lib/installer.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+
+    def _install_npm(self):
+        self._run_command(
+            'npm install -g',
```
Yajo: Add [npm cleanup](https://github.com/Tecnativa/docker-odoo-base/blob/master/Dockerfile#L102) after this.

lib/installer.py
```diff
@@ -0,0 +1,98 @@
...
+        if split:
+            command = command.split()
+        for requirement in self.requirements:
+            self._run_command(
+                command + [requirement],
+                split=False,
+                shell=shell,
+            )
+
+    def _run_command(self, command, split=True, shell=False):
```
Yajo: Didn't you add a helper for this already in odoobaselib?

lasley:
>  I mean that, after all, this code you are adding is not meant to be reused anywhere, so no need to add classes and so on; you can just add some helper methods to the odoobaselib, and then do a simple, hardcoded, sequential script that just makes its checks and installs everything, just like you would do with bash.

I've already reused the class in #28 & it's used in the removal script too. IMO at the point three scripts are using the same context, a class was a good idea.

lib/installer.py
lasley: We can, however I found it difficult in my debugging to find which packages were failing. If we try to install everything at once, installation issues have to be handled one at a time. If we separate and install, all the issues can be identified and resolved after the first build or so.

lasley:
Rebase complete & PR comments attended to (except https://github.com/Tecnativa/docker-odoo-base/pull/33#discussion_r122649706). Note that I squashed all existing commits into 3ed2d8e in order to make the rebase easier. The commit responding to the most recent review is 	dc5c4c8

README.md
lasley: Ooops - I never actually did this

Yajo: I did it for you, don't worry 😆

Issue #37, opened 2017-06-20
[RFC] Move config generation to entrypoint


lasley:
I wanted to open a discussion to move the Odoo configuration generation from the build into the entrypoint. I am mainly wondering what were the design considerations that led you to put it in the build?

My reasoning for entrypoint is as follows:
* Passwords are currently hard coded into the build image
  * Security issue - passwords should not be stored in the container registry (note that [ECR is not a HIPAA compliant service](https://aws.amazon.com/compliance/hipaa-eligible-services-reference/))
  * Maintenance issue - our policies require that all passwords are rotated on a consistent basis. This is much more difficult at the image level
  * Passwords stored in repo unencrypted (unless I'm missing something)
* Exponential image duplication - It is possible to have multiple customers on the same image if the config is generated on the fly
  * This is particularly important in my SaaS, where all images are identical except for the credentials.

Ummm that's all I can think of at the moment. Any thoughts?

pedrobaeza:
I don't see special problem right now, although I'm not an specialist on Docker, but how do you plan to give passwords then?

@Yajo only works on mornings, so you won't receive his answer till tomorrow.

lasley:
> but how do you plan to give passwords then?

Environment variables. This is the same strategy used by the Postgres container, which takes `POSTGRES_PASSWORD` + `POSTGRES_USER` env vars & creates the role based upon them on initial launch. 

In our case, I think we'd be safe generating the config on the fly every time as opposed to the init that the Postgres one does. The Postgres init is fairly high overhead, vs we're just text editing.

Taking it a step further, we have another repo for our customer deploys. This repo consists of folders for the customers, subfolders for their instances. In the subfolders are two files:
* `docker-compose.yml.vault`
* `rancher-compose.yml` 

These files describe the specifics of the instance deploy, with the `docker-compose` file being encrypted with Ansible vault. Our Ansible deployer unvault-ifies it & triggers the Rancher deploy.

> @Yajo only works on mornings, so you won't receive his answer till tomorrow.

No worries - Admittedly I had already kind of figured out his schedule at this point. Lucky duder - wish I worked only mornings!

lasley:
It should also be noted that we're maintaining different vault passwords for each customer subfolder. I'd be happy to discuss that one in person, but it's a bit in depth & I'm not sure how public I want it just yet - for all I know, there's a gaping security hole 😉

pedrobaeza:
[...]

Issue #20, opened 2017-03-21
Move scaffolding to Docker Swarm


Yajo:
WIP of moving scaffolding to docker-compose file v3.1 and supporting only Docker Swarm (>=1.13) in `prod.yaml` can be found in https://github.com/Tecnativa/docker-odoo-base/tree/scaffolding-composefile3.

lasley:
IMO it feels a bit early to deprecate the older compose files. This is just scaffold though, right? So really it's just a sample, and the user is free to use whatever?

Yajo:
Yeah, [of course](https://github.com/Tecnativa/docker-odoo-base#can-i-have-my-own-scaffolding).

There's still https://github.com/docker/compose/issues/4315, and I'm also considering moving the scaffolding to make use of [ansible-container](http://docs.ansible.com/ansible-container/) instead of plain docker-compose.

Still nothing decided, but keep an eye on this issue for progress 😉

lasley:
Ahahahah I'm so happy you guys are loving Ansible!

Yajo:
OK, now there's a way to define different environments with v3 of docker compose file: https://github.com/moby/moby/issues/31101#issuecomment-330152093 (it was there before, I just didn't read carefully enough).

That would count as a breaking change, but allows to move for v3. [Still I'm not sure if ansible would support having v3](http://docs.ansible.com/ansible/latest/docker_service_module.html#synopsis), which is very important for us too...

lasley:
I still think this move is quite early personally. Also, supporting only Swarm seems like cutting out a good amount of other uses out of the picture. We use Kubernetes here, and likely won't be switching any time soon. 

Ansible - yeah it uses `docker-py` if you're using the `docker` module, which isn't compatible with `docker-compose`. People on the interwebz look to be using docker shell commands when newer syntax is required.

Yajo:
Well, switching to docker compose file v3+ does not force swarm, it just enables it. You can still use swarmless docker with that file version.

Yajo:
Will do later in https://github.com/Tecnativa/doodba-scaffolding

PR #38, opened 2017-06-20
[IMP] Move config generation to entrypoint

lasley:
I'm pretty sure this is the only change required for #37, although I'll test later today. 

I still need to update the docs and all that, but I figured I'd put my code where my mouth is - particularly seeing as I need it this way anyways 😆 

I left the possibility of generating the config at build so that we don't kill your images, although I recommend we leave the default to build on entrypoint instead. This means that you have to edit all your images though, which kind of sucks, but the other way is insecure by default which I think is bad.


lasley:
Seems this works too - hooray!

build.d/60-config-generate
```diff
@@ -1,9 +1,8 @@
...
-set -e
-src="/opt/odoo/common/conf.d/* /opt/odoo/custom/conf.d/*"
-log INFO Merging $(ls $src | wc -l) configuration files in $OPENERP_SERVER
-conf=$(cat $src | envsubst)
-log DEBUG "Resulting configuration:
-$conf"
-echo "$conf" > $OPENERP_SERVER
+
+if [ "${CONFIG_BUILD}" != true ]; then
+    log WARNING Skipping Odoo config generation - will perform during entrypoint.
```
Yajo: If this is the new normal behavior, then the warning should go in the opposite case.

entrypoint.d/50-config-generate
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [ "${CONFIG_BUILD}" = true ]; then
```
Yajo: This env will never be available at run time (notice it is an `ARG`, not an `ENV` in the Dockerfile).
Instead, you can just check for existence of the file indicated in `$OPENERP_SERVER`.

lasley:
This looks to be working in my tests

Yajo:
But one question... when you reboot Odoo, do you want a config file to be regenerated? I think that would make most sense, and in such case, we should try to keep it simple and remove the thing from build and move it to entrypoint, no less, no more.

lasley:
I agree we want it generated every time. Honestly the only reason I left it in build was for backwards compatibility, but if you're good with not supporting it so am I. All my scaffolds use environment vars, so it doesn't affect me really.

Dockerfile
```diff
@@ -16,6 +16,7 @@ ONBUILD ARG DEPTH_DEFAULT=1
 ONBUILD ARG DEPTH_MERGE=100
 ONBUILD ARG CLEAN=true
 ONBUILD ARG COMPILE=true
+ONBUILD ARG CONFIG_BUILD=false
```
Yajo: What's this for then?

entrypoint.d/50-config-generate
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+config-generate
```
Yajo: Why not just put the file here? I see no more use cases...

Yajo:
I like it now 😊 

Can you fix conflicts and squash please?

lasley:
@Yajo done

Yajo:
This one broke many things, I reverted it at #48.

Sorry to say, but merging nothing more until #43 is fixed. 😞

lasley:
@Yajo - what did it break? I had already warned that essentially all of your scaffolding and existing images would be broken with this in #37 & it seemed you guys were good with this.

As it stands now, this repo cannot scale past 1:1 image:customer. This is an absolute blocker for use of this repo, as well as a huge security issue storing the passwords for the database and instance admin passwords in the images.

Yajo:
Yeah, it's not your fault, it's mine for merging without proper testing, but I had to roll it all back in a rush. 😅 

We want this feature, but cannot merge without proper backwards compatibility.

The main problem was that older deployments had build args where new ones would need env vars. Build args do not get converted into env vars unless explicitly made in the Dockerfile. That's most likely the backwards compatibility layer that we will need to maintain for awhile. Something like:

```Dockerfile
ONBUILD ARG WITHOUT_DEMO=all
ONBUILD ENV WITHOUT_DEMO="$WITHOUT_DEMO"
```

... the same as that for every variable used under `common/conf.d`. This way if you use env vars, default ones get overriden at runtime through your docker-compose file, but if you have an old deployment using build args, those get converted into env vars at build time that can be used at runtime to get the config file generated.

As you can see, not much hassle, but I had no time to fix & test properly and I had to roll it back or we were going to get broken builds everywhere.

Let's just reopen this in a new PR and continue from there if you want. 😉 

> As it stands now, this repo cannot scale past 1:1 image:instance.

I don't understand what you mean...

> huge security issue storing the passwords for the database and instance admin passwords in the images.

Unless you use secrets, it's not much different from storing them in runtime, but using secrets involves fixing #20, which is in the roadmap, but is not our current utmost priority.

lasley:
> As it stands now, this repo cannot scale past 1:1 image:instance.
>> I don't understand what you mean...

The Docker images that are built contain an admin and a Postgres password. Unless you plan on sharing passwords between your customer instances, this is a bottleneck to vertical scaling. IMO database name & user is too.

> Unless you use secrets, it's not much different from storing them in runtime, but using secrets involves fixing #20, which is in the roadmap, but is not our current utmost priority.

It's much different. My container infrastructure is all in HIPAA compliant, audited infrastructure. Build images are stored either in Docker or AWS Container Registry. Neither of these services are secure, and thus my passwords have been exposed across a known insecure plane.

Yajo:
OK now I get your points, thanks for the explanation. So now the goal is to fix #43 and then add this feature again 😊

PR #39, opened 2017-06-20
[IMP] Change Odoo user shell to `/bin/false`

lasley:
I propose changing the Odoo user's shell to `/bin/false`. Relevant conversation - https://github.com/Tecnativa/docker-odoo-base/commit/3256bc955b50cf277142dd76e633bdca7fcb96e1#commitcomment-22638986

If the quick `su` access is required, maybe we could use a bash alias in the root user?


Yajo:
I wasn't aware of `-s` flag in `su`, so of course you're right here; merging.

PR #41, opened 2017-06-20
[IMP] Add automatic installation of Odoo addons

lasley:
This adds automatic installation of addons that are defined in the `addons.yml` file. It also adds a nifty method to wait on containers to exit, which can be used to allow an install container to be launched before other ones with a `WAIT_CONTAINER` env var as can be seen in my test config:

```yml
services:

 install:
    image: lasley/odoo-buildouts:laslabs-10
    depends_on:
      - db
    tty: true
    environment:
      ADDON_INSTALL: 'true'

 web:
    image: lasley/odoo-buildouts:laslabs-10
    restart: unless-stopped
    depends_on:
      - db
      - install
    tty: true
    ports:
      - 8069
      - 8071
    environment:
      WAIT_CONTAINER: 'install'

  db:
    image: postgres:9.6-alpine
    restart: unless-stopped
    depends_on:
      - db-utils
    environment:
      PGDATA: /var/lib/postgresql/data/pgdata
      POSTGRES_PASSWORD: 'derp'
      POSTGRES_USER: 'odoo'

  db-utils:
    image: grupocitec/pgutils
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,6 +54,7 @@ ARG PYTHONOPTIMIZE=2
 ARG WKHTMLTOPDF_VERSION=0.12.4
 ARG WKHTMLTOPDF_CHECKSUM='049b2cdec9a8254f0ef8ac273afaf54f7e25459a273e27189591edc7d7cf29db'
 ENV OPENERP_SERVER=/opt/odoo/auto/odoo.conf \
+    WAIT_NOHOST='' \
     UNACCENT=true \
     # Git and git-aggregator
     GIT_AUTHOR_NAME=docker-odoo \
--- a/README.md
+++ b/README.md
@@ -315,6 +315,44 @@ now keep this in mind:
 The CLI text editor we all know, just in case you need to inspect some bug in
 hot deployments.
 
+### `install-addons`
+
+This script will iterate the `addons.yml` file to gather support addons, and
+subsequently install to the Odoo instance. It will then exit.
+
+This script is great to use during buildout in order to prepare an instance.
+Below is a minimalistic `docker-compose` file illustrating a use case:
+
+```yaml
+version: '2'
+
+services:
+
+  install:
+    context: ./odoo
+    command: 'install-addons'
+    tty: true
+    depends_on:
+      - db
+
+  web:
+    context: ./odoo
+    restart: unless-stopped
+    tty: true
+    depends_on:
+      - db
+      - install
+    ports:
+      - 8069
+      - 8071
+    environment:
+      WAIT_NOHOST: 'install'
+
+  db:
+    image: postgres:9.6-alpine
+    restart: unless-stopped
+```
+
 ### `log`
 
 Just a little shell script that you can use to add logs to your build or
--- a/entrypoint.d/999-odoo-wait
+++ b/entrypoint.d/999-odoo-wait
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+if [ -z "${WAIT_NOHOST}" ]; then
+    exit 0
+fi
+
+log INFO "Waiting for the Odoo in ${WAIT_NOHOST} to stop"
+
+while ping -c1 "${WAIT_NOHOST}" &> /dev/null; do
+    log DEBUG "${WAIT_NOHOST} responded to ping"
+    sleep 1
+done
+
+log INFO "${WAIT_NOHOST} has stopped responding"
+
+exit 0
--- a/lib/odoobaselib/__init__.py
+++ b/lib/odoobaselib/__init__.py
@@ -34,6 +34,12 @@ else:
 logging.root.setLevel(_log_level)
 
 
+def addons_active():
+    """Returns a list of active addons, as defined in `addons.yml`."""
+    raw_addons = sum(addons_config().values(), [])
+    return filter(lambda x: x != '*', raw_addons)
+
+
 def addons_config():
     """Load configurations from ``ADDONS_YAML`` into a dict."""
     config = dict()
--- a/bin/install-addons
+++ b/bin/install-addons
@@ -1,0 +1,9 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from odoobaselib import addons_active, do_command
+
+
+do_command(
+    'odoo --stop-after-init -i %s' % ','.join(addons_active())
+)


lasley:
Reporting back - seems it works!

I am having trouble exiting at the end though. At the moment I'm just using `exit 0` in the bash script that has a `+e` in it, but I'm getting a not found error:

```
odoo-install_1  | + exec exit
odoo-install_1  | /opt/odoo/common/entrypoint.sh: line 20: exec: exit: not found
```

It does work though, because the bad command causes the script to exit. I'd just like something a bit cleaner - so was wondering what the typical strategy is for this?

pedrobaeza:
Can we blacklist some of the addons of the list? Example: I want to have available the module web_environment_ribbon, but don't install it on the production database.

lib/odoobaselib/__init__.py
```diff
@@ -35,6 +35,12 @@
 logging.root.setLevel(_log_level)
 
 
+def addons_active():
+    """ Returns a list of active addons, as defined in `addons.yml`. """
```
Yajo: Please remove the surrounding spaces in the docstring, following [PEP-0257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings).

entrypoint.d/999-odoo-wait
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+if [ "${WAIT_CONTAINER}" == false ]; then
+    exit 0
+fi
+
+log WARNING "Waiting for '${WAIT_CONTAINER}' to stop"
+
+while ping -c1 "${WAIT_CONTAINER}" &> /dev/null; do
+    log INFO "${WAIT_CONTAINER} responded to ping - waiting for exit."
```
Yajo: This should be `DEBUG`, or your logs can get full of these...
Also, I'd add a `sleep 1` so we wait 1s among pings.

entrypoint.d/999-odoo-wait
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+if [ "${WAIT_CONTAINER}" == false ]; then
+    exit 0
+fi
+
+log WARNING "Waiting for '${WAIT_CONTAINER}' to stop"
```
Yajo: This should be an INFO IMHO.

entrypoint.d/999-odoo-wait
```diff
@@ -0,0 +1,15 @@
...
+
+while ping -c1 "${WAIT_CONTAINER}" &> /dev/null; do
+    log INFO "${WAIT_CONTAINER} responded to ping - waiting for exit."
+done
+
+log WARNING "${WAIT_CONTAINER} has stopped responding - continuing boot."
```
Yajo: INFO too.

Dockerfile
```diff
@@ -43,6 +43,8 @@ ARG PYTHONOPTIMIZE=2
 ARG WKHTMLTOPDF_VERSION=0.12.4
 ARG WKHTMLTOPDF_CHECKSUM='049b2cdec9a8254f0ef8ac273afaf54f7e25459a273e27189591edc7d7cf29db'
 ENV OPENERP_SERVER=/opt/odoo/auto/odoo.conf \
+    ADDON_INSTALL=false \
```
Yajo: The WAIT_CONTAINER stuff is OK, but drop this `ADDON_INSTALL` feature. It makes the image more complex, and after all you can do the same replacing your:

```yaml
installer:
  environment:
    ADDON_INSTALL: "true"
```

by:

```yaml
installer:
  command: install-addons
```

... so no real benefit, unless I'm missing something... 🤔

Dockerfile
```diff
@@ -43,6 +43,8 @@ ARG PYTHONOPTIMIZE=2
...
+    WAIT_CONTAINER=false \
```
Yajo: What if you container is called `false`? OK, it's stupid, but since this is not a boolean value, I prefer to keep it as an empty string/filled string.

Also, call it `WAIT_NOHOST`, since it could be any host that can be resolved, and it is supposed to be unavailable for odoo to start.

lib/odoobaselib/__init__.py
```diff
@@ -35,6 +35,12 @@
...
+    raw_addons = list(reduce(lambda x, y: x + y, addons_config().values()))
+    return list(filter(lambda x: x != '*', raw_addons))
```
Yajo: Remove `list`

lib/odoobaselib/__init__.py
```diff
@@ -35,6 +35,12 @@
 logging.root.setLevel(_log_level)
 
 
+def addons_active():
+    """ Returns a list of active addons, as defined in `addons.yml`. """
+    raw_addons = list(reduce(lambda x, y: x + y, addons_config().values()))
```
Yajo: `raw_addons = sum(addons_config().values(), [])`

lasley:
@pedrobaeza - My plan admittedly was to pair the wildcard with this in order to force install specific addons while advertising the rest. My use case is the SaaS though, where modules-a-plenty is kinduv the goal. 

What do you propose? It sounds to me like you wouldn't want to advertise this module to your production instance either though, right? Are there other use cases where you want to selectively advertise, but not manage the installation of, an addon?

entrypoint.d/999-odoo-wait
```diff
@@ -1,15 +1,16 @@
 #!/bin/bash
 
-if [ "${WAIT_CONTAINER}" == false ]; then
+if [ "${WAIT_NOHOST}" == "" ]; then
```
Yajo: `if [ -z "$WAIT_NOHOST"]; then`

README.md
```diff
@@ -317,6 +317,45 @@ now keep this in mind:
...
+
+services:
+
+  install:
+    image: laslabs/odoo-buildouts:dev-10
+    command: 'install-addons'
+    tty: true
+    depends_on:
+      - db
+    tty: true
```
Yajo: tty twice

README.md
```diff
@@ -317,6 +317,45 @@ now keep this in mind:
...
+This script is great to use during buildout in order to prepare an instance.
+Below is a minimalistic `docker-compose` file illustrating a use case:
+
+```yaml
+version: '2'
+
+services:
+
+  install:
+    image: laslabs/odoo-buildouts:dev-10
```
Yajo: use:

        build:
            context: ./odoo

Remember that this project is tightly coupled with its scaffolding, adding more unknown images as this to the examples adds confusion and diverges from the original purpose.

(I'm not saying your approach is wrong, but just want to keep the project consistent 😊 )

You can grab any docker-compose.yaml samples you need from the combination of [common.yaml](https://github.com/Tecnativa/docker-odoo-base/blob/scaffolding/common.yaml) and [prod.yaml](https://github.com/Tecnativa/docker-odoo-base/blob/scaffolding/prod.yaml)

README.md
```diff
@@ -317,6 +317,45 @@ now keep this in mind:
...
+    command: 'install-addons'
+    tty: true
+    depends_on:
+      - db
+    tty: true
+
+  web:
+    image: laslabs/odoo-buildouts:dev-10
```
Yajo: same here

README.md
```diff
@@ -317,6 +317,45 @@ now keep this in mind:
...
+      - db
+      - install
+    ports:
+      - 8069
+      - 8071
+    environment:
+      WAIT_NOHOST: 'install'
+
+  db:
+    image: postgres:9.6-alpine
```
Yajo: `postgres:${DB_VERSION}-alpine`

Dockerfile
```diff
@@ -111,7 +112,7 @@ RUN curl -SLo wkhtmltox.tar.xz https://github.com/wkhtmltopdf/wkhtmltopdf/releas
 WORKDIR /opt/odoo
 RUN pip install --no-cache-dir \
     astor git-aggregator openupgradelib ptvsd==3.0.0 pudb wdb
-COPY bin/autoaggregate bin/install.sh bin/log bin/pot bin/python-odoo-shell bin/unittest /usr/local/bin/
+COPY bin/* /usr/local/bin/
 COPY bin/direxec.sh common/entrypoint.sh
 RUN ln common/entrypoint.sh common/build.sh
```
Yajo: If you change to copy `bin/*`, then change these 2 lines to:

```Dockerfile
RUN ln -s /usr/local/bin/direxec.sh common/entrypoint.sh \
    && ln -s /usr/local/bin/direxec.sh common/build.sh
```

entrypoint.d/999-odoo-wait
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+if [ "${WAIT_NOHOST}" == "" ]; then
```
Yajo: -z "$WAIT_NOHOST"

README.md
lasley: I feel like this adds complexity that is not required for an example. For me- the environment variables that are present on build took a long time to understand in terms of what needs to be present at each build step, which only got confused further when I started thinking mass distribution of images.

For example, my `docker-compose.yml` for builds look like this:

```yml
version: "2.1"

services:
  odoo:
    image: lasley/odoo-buildouts:smd-10
    build:
      context: ./odoo
```

Really it could be a simple `docker build` if I wasn't lazy and didn't want to have to remember the image name. 

I digress though, I think static version is better here.

pedrobaeza:
@lasley the use case can be generalized as the following: having modules that are installed on demo instance, but not on production. As the addons file is shared between both, you can't differentiate it.

lasley:
@pedrobaeza - No problem on the use case, I was more asking how you propose this be implemented in the yml file? For a blacklist we would change the addons syntax to something like the below?

```yml
web:
- web_responsive
- web_environment_ribbon:no-install
```

pedrobaeza:
Uhm, I'm thinking that this can be interesting to be done selectively by environment (prod, test, devel...), because you can specify ribbon module to be auto-installed on demo but not on production, but I don't know if we have any variable for knowing that. The syntax can be:

```yml
web:
    - web_responsive
    - web_environment_ribbon
        - env: demo,devel
```
If no env is specified, then it's installed everywhere. If there's any, it's only applied on that list.

lasley:
I don't know of a variable that knows that either. @Yajo - do we have anything like this that we're missing?

README.md
Yajo: Right, but the problem is that I will have to maintain the example, because if at some point the client libraries get upgraded, it could work no more.

I imagine an issue saying: "I followed this example and it does not work"; me looking at the example, it's wrong, update version...

That variable is self-descriptive enough to make the example reader what it does, so please use it.

Yajo:
Currently the built image has no notion of the environment it runs. However one can read the odoo.conf and see if we have the demo data enabled or not. That can be a good indicator of wether we are in production or not.

Also, AFAIK [this syntax](https://github.com/Tecnativa/docker-odoo-base/pull/41#issuecomment-310688408) is not possible.

lasley:
I went ahead and rebased this one.

Regarding the selective installation of module dependent on environment - what is the use case on this? I'm not sure if just targeting on the demo data would be clear enough, hence me trying to understand a bit more.

lasley:
NVM you gave a use case - web_environment_ribbon. 

Hmmm so I think the real solution would be some sort of run-parts implementation on the addons yml, with configuration based on environment variables. Anything else just seems kind of hacky I think - particularly because we'd have to change up the syntax of the pre-existing addons file.

Yajo:
I'm thinking that [YAML already supports multiple docs in single file](http://www.yaml.org/spec/1.2/spec.html#id2760395). The new syntax, which would be almost completely backwards-compatible, could be:

```yaml
# Spanish Localization
l10n-spain:
    - l10n_es
server-tools:
    - date_range
---
# SEO tools
website:
    - website_blog_excertp_img
server-tools: # Here we repeat server-tools, but no problem because it's a
              # different document
    - html_image_url_extractor
    - html_text
---
# Above blocks work as they do until today.
# Below block adds the special ONLY key, with env variable matches
ONLY:
    PGDATABASE:
        # it can be a list
        - devel
        - test
web:
    - web_environment_ribbon
---
# This is only for prod databases
ONLY:
    PGDATABASE: prod  # it can be a string
server-tools:
    - auto_backup
```

PR #42, opened 2017-06-21
[FIX] SSH directory should be private

lasley:


Dockerfile
```diff
@@ -35,8 +35,10 @@ ONBUILD ARG LOCAL_CUSTOM_DIR=./custom
 ONBUILD COPY $LOCAL_CUSTOM_DIR /opt/odoo/custom
 # https://docs.python.org/2.7/library/logging.html#levels
 ONBUILD ARG LOG_LEVEL=INFO
+ONBUILD RUN ln -s /opt/odoo/custom/ssh ~root/.ssh \
+            && chmod -Rc 0600 /opt/odoo/custom/ssh/* \
+            && chmod -c 0700 /opt/odoo/custom/ssh/
```
Yajo: - move this to common/build.d/80-permissions.
- do not use octal permissions, they are less readable than using letters.
- a single `chmod -Rc u=rwX,go= custom/ssh` would do the same
- the `ln -s` part should be moved to the base image (outside an `ONBUILD`) because symlink creations don't need an existing target to success, so we drop one layer from the onbuild and move it to base image. (This was my fault, but now that you're fixing it... 😇 )

lasley: > move this to common/build.d/80-permissions.

Can't happen - this has to be performed before the build. Otherwise it's not used for the repo aggregation. The other permissions have to happen after the repo aggregation, so it's a rock and a hard place.

> the ln -s part should be moved to the base image (outside an ONBUILD) because symlink creations don't need an existing target to success, so we drop one layer from the onbuild and move it to base image. (This was my fault, but now that you're fixing it... 😇 )

We would still have the same amount of layers, because the permissions changes are still required.

lasley:
Looks like this one works still too!

We do have an alternative to the permissions in the Dockerfile, which would be to create another permissions script that is explicitly executed before all build scripts. I feel like creates some ambiguity between the other permissions script, and also feels kind of like Ruby on Rails magic if we bury it into an explicit something somewhere. 

Thoughts? Ideaz?

Yajo:
Ah OK I get it now.

Then you should take care of this in `80-permissions` to make him not touch the `ssh` folder (we don't want it back again with wide permissions, right?)

lasley:
Good call, @Yajo

PR #43, opened 2017-06-23
docker-compose based CI

Yajo:
let's see if I understood Travis...

This seems more what I wanted for CI here.


lasley:
Ok so I can see you really really want docker-compose. I'll see if I can get that added into our system this weekend 😉

tests/basic/custom/tests/0010-odoo-install-all
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+set -e
+odoo --stop-after-init --install all
```
pedrobaeza: Is this going to install all Odoo addons? Is the intended behavior?

lasley:
@Yajo - FYI you're fighting against [this](https://github.com/LasLabs/docker-quality-tools/blob/master/travis/travis_run_test), which runs the Dockerfile.

Yajo:
I don't want to fight it, but AFAICS it is not designed with docker-compose in mind 😅 

This is a wip, I just opened it to let you know what was on my mind (I see it worked! 😄). If we can reuse docker-quality-tools but still do docker-compose based tests, it'll be OK for me.

The idea is that we can imitate the normal workflow for this project, which is using a clone of the scaffolding branch, changing it in some ways, and using docker-compose to start the project in some environment. You see, this image is a base for subimages, so if we don't have a subimage to test, the tests will not be complete.

To give you an example, last week when I merged #42, I had a problem in old projects that had not included https://github.com/Tecnativa/docker-odoo-base/commit/1b0d4a6d05904bf6f11ea2aef319bf7c8d09a5ef, so I had to fix it in https://github.com/Tecnativa/docker-odoo-base/commit/bd61df6ceb5107c66ad467484953e974ef285ebb. Then I had to add another fix to the scaffolding (https://github.com/Tecnativa/docker-odoo-base/commit/b0440d651d3d054fe218096e290556c097a33bfb) to make the setup-devel environment work again. Unless we imitate all that workflow in the tests, they will not be of trust IMO.

Do you think you can PR to this branch?

lasley:
Yeah TBH I designed DQT before I knew the power of Docker Compose. You can see that I basically redesigned a few pieces of it too.

Why do we have to use compose though? I'm using this repo, including scaffolding, without compose. Would you accept a PR using the scaffold branch instead of the test ones, but without implementing the docker-compose workflow?

Yajo:
This project's goal is to supply a plug&play odoo deployment solution. It relies on some tools (git-aggregator, some build scripts, and docker-compose...) and some decisions (repos.yaml, addons.yaml, custom folder structure, scaffolding, separation of 3 environments...).

If you really don't want to use docker-compose, of course you can still make this work, as if you prefer to install addons through pip, or if you prefer to create a script in `build.d` that clones all you need by git instead of using git-aggregator, or if you prefer git submodules, or if you prefer to put odoo.conf in another location or run a different command... Of course you can use it, but the point is that's not our *supported* way of using the tool.

The supported way is, basically, to follow instructions in README.

Having just one supported workflow gives many benefits to us:

- Training newcomers is much easier: Just do A - B - C and it will work. Further knowledge can come later with further needs.
- Newcomers get trained to use standard tools. We don't try to *hide* docker-compose by creating a list of aliases to it or scripts that do somehow what it does... We just use a standard tool that you will be able to continue using for many years.
- Newcomers to the docker world will almost never have to run `docker` directly; with `docker-compose` should be enough for most cases, which is a ton easier.
- Hardcoded directory structure lets everyone know where everything should be, so if you search for an addon, once you look for it once, you will always know where the next must be, and so on.
- Your git project envolves everything: external referencies to addons, required packages for any installer method, working definitions of all needed environments... Just run it and it will work. If anything is missing, add it and it will work on all environments out of the box.

So yes, tests without compose could be accepted, but still not as our *main* tests. Our main goal with tests is to test our supported workflow first, and then maybe others later.

I hope you understand...

BTW, I encourage you to try following the supported workflow based on the combination of docker-compose + separated environment files + scaffolding, at least for just one week. 😉  I'm not sure if you will be able to grasp the project's spirit unless you follow the workflow... 🤔 , and OTOH the workflow is not perfect, it's evolving too, so help on the workflow can be welcome too.

lasley:
@Yajo - I get where you're coming from, the problem is that there are essentially no build args that I need to add for a default production installation - assuming #38. Maybe an Odoo/Postgres version here or there, but really most everything is in the entrypoint.

The huge problem with your one supported workflow, and why it is entirely incompatible for me (believe me, I tried the hell out of it), is that I cannot build images at my deploy layer. This means that a separate build system is required. In the world of medical & the fact that the password to the database is stored in the image, my build system has to be HIPAA compliant - which requires a dedicated EC2 instance at a minimum of m4.large size + a complete STIG rollout with compliant logging. 

Additionally, it is assumed that every system is compatible with an `env` file. This is also not the case, an example system being Rancher.

Another assumption is that each image is dedicated to one production instance, which is horizontal scaling. My goal is to cover a few hundred production instances with maybe 6-7 Docker images to maintain.

But really this was just me explaining my use case. I never expected that to become the supported method, and I'm honestly good with it being obscured so that my competitors can't do it. If you're curious, [this is a scaffold that I put together real quickly to illustrate to you what I'm going for](https://github.com/LasLabs/docker-odoo-medical/tree/feature/vertical-medical/10.0).

That said, I'm in the middle of a crapton of production deploys at the moment and likely will not have time to get to this. It sucks that we're at a feature freeze though - I'm going to have to develop in a forked incubator then; there's still more things I need 😦

pedrobaeza:
We can get to an understanding of course. The real thing is that your requirements don't break ours. Today we had broken builds because of the merged entrypoint system, and without time for checking what is the problem.

If you think the best way is to have an incubator for using your features immediately and propose meanwhile here with more calm, then go ahead.

lasley:
>  The real thing is that your requirements don't break ours. Today we had broken builds because of the merged entrypoint system, and without time for checking what is the problem.

I totally warned about this though, and even included a system to allow them to be compatible. I was told to take it out.

pedrobaeza:
Yeah, I know, but it was more a bug that a change in the scaffolding. I don't have the details now, but tomorrow we will check and restore the functionality if it's easy to fix.

lasley:
Please - I'd love to know what was wrong, if not just for intellectual curiosity.

Yajo:
@lasley sorry for disturbance. I understand that having a bunch of rollbacks on your features can be somehow offending, but that's not my intention here. We're very grateful to have your contributions and we really wish to be able to have a common base to work with.

We just had to do that in a hurry because of the broken workflow (I just explained in another thread).

> I'm going to have to develop in a forked incubator then; there's still more things I need 😦

Please do! There's no problem on that 😊 

I think it's the best approach. You and us have mostly the same goal but different priorities right now. Ours currently (with this project) is to fix this PR and have proper testing, so we can continue evolving from there on.

But I think the best you can do is to have your own fork for now and open the needed PR here, while you can keep them merged in your master. We don't want to block you, but we need a little time to do things the right way, and I'm sure you agree 😉 

Eventually we'll have proper CI, and we will be able to start testing and merging everything and we will be able to share the same common base again. 😊 

Regards

PR #51, opened 2017-07-18
fix UID warning for systems with readonly UID

yucer:
Some linux systems don't allow to modify UID as variable. When used in
.bashrc the warning appears all the time.

After this change. use instead:

     export DEV_UID="$(id -u $USER)" GID="$(id -g $USER)" UMASK="$(umask)"

### DIFF
--- a/setup-devel.yaml
+++ b/setup-devel.yaml
@@ -25,7 +25,7 @@ services:
         environment:
             DEPTH_DEFAULT: 100
             # XXX Export these variables before running setup to own the files
-            UID: "$UID"
+            UID: "$DEV_UID"
             GID: "$GID"
             UMASK: "$UMASK"
         user: root


Yajo:
Have you tried by just doing this?:

    $ export UID

Yajo:
OK I have made some tests and it seems like it's Bash that treats that variable as a readonly variable. If you try with sh or fish, you'll see there is no problem. I currently use fish, that's why I did not realize of this before.

However, in my tests, by doing https://github.com/Tecnativa/docker-odoo-base/pull/51#issuecomment-316293838 it seems to work with no problem. Could you test? If so, I'd prefer to add a FAQ section instead.

Thanks!

yucer:
It works. 

The exact error was:

    bash: UID: readonly variable

In case you want to add it to the FAQs.

Thanks.

PR #52, opened 2017-07-18
[FIX] Create directories that are expected to exist

lasley:
* Create ssh directory
* Create private addons directory

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ ONBUILD COPY $LOCAL_CUSTOM_DIR /opt/odoo/custom
 ONBUILD ARG LOG_LEVEL=INFO
 ONBUILD RUN mkdir -p /opt/odoo/custom/ssh \
             && ln -s /opt/odoo/custom/ssh ~root/.ssh \
-            && (chmod -R u=rwX,go= /opt/odoo/custom/ssh || true)
+            && chmod -R u=rwX,go= /opt/odoo/custom/ssh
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -35,7 +35,8 @@ ONBUILD ARG LOCAL_CUSTOM_DIR=./custom
 ONBUILD COPY $LOCAL_CUSTOM_DIR /opt/odoo/custom
 # https://docs.python.org/2.7/library/logging.html#levels
 ONBUILD ARG LOG_LEVEL=INFO
-ONBUILD RUN ln -s /opt/odoo/custom/ssh ~root/.ssh \
+ONBUILD RUN mkdir -p /opt/odoo/custom/ssh \
+            && ln -s /opt/odoo/custom/ssh ~root/.ssh \
             && (chmod -R u=rwX,go= /opt/odoo/custom/ssh || true)
```
Yajo: This `|| true` can disappear now then

tests/__init__.py
```diff
@@ -96,6 +96,7 @@ def test_smallest(self):
         commands = (
             # Must generate a configuration file
             ("test", "-f", "/opt/odoo/auto/odoo.conf"),
+            ("test", "-d", "/opt/odoo/custom/src/private"),
```
Yajo: Please add also `test -d custom/ssh`, and remove the `ssh/.empty` file (to ensure that it will exist even if your current scaffolding has no `ssh` dir)

PR #56, opened 2017-07-25
[FIX] Switch Odoo version and source args to onbuild

lasley:
From what I can tell, this is required in order to properly install a different Odoo version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,9 +127,9 @@ RUN chmod -R a+rx common/entrypoint* common/build* /usr/local/bin \
 # Execute installation script by Odoo version
 # This is at the end to benefit from cache at build time
 # https://docs.docker.com/engine/reference/builder/#/impact-on-build-caching
-ARG ODOO_SOURCE=OCA/OCB
-ARG ODOO_VERSION=10.0
-ENV ODOO_VERSION="$ODOO_VERSION"
+ONBUILD ARG ODOO_SOURCE=OCA/OCB
+ONBUILD ARG ODOO_VERSION=10.0
+ONBUILD ENV ODOO_VERSION="$ODOO_VERSION"
 RUN install.sh
 
 # Metadata


Yajo:
This part is essentially the one that takes care of producing different v8, v9 and v10 image versions with their corresponding tags, and make all of them share all possible layers. That process is handled by the scripts found at [hooks](https://github.com/Tecnativa/docker-odoo-base/tree/master/hooks).

It configures [the `install.sh` script](https://github.com/Tecnativa/docker-odoo-base/blob/master/bin/install.sh) that goes just next, which install everything that should be Odoo-flavor agnostic (basically, [it gives it a place where to look for the `requirements.txt` file](https://github.com/Tecnativa/docker-odoo-base/blob/master/bin/install.sh#L4)).

The real Odoo code is downloaded, merged and installed later [in the `build.d` scripts](https://github.com/Tecnativa/docker-odoo-base/blob/master/Dockerfile#L41), but can be downloaded and merged in your favourite flavor by using `repos.yaml`.

So, basically, if you need to install a different odoo version, just base your project on `tecnativa/odoo-base:$ODOO_VERSION` and continue as usual from there. 😉 

Closing because this would break everything and I feel the PR comes more from a misunderstanding than from anything else. Feel free to reopen if needed 😊

lasley:
Thanks for the explanation @Yajo - I noticed that for the most part I was able to simply use the v10 image with v9 repos & make the symlink myself.

Why have multiple versions of prebuilt images without code? Aren't the binary dependencies between Odoo versions the same? Maybe there will be a few more in the v10, but I don't think it would hurt a v9?

Also - I can't use your images, have to use my own because I need the config generation on the fly.

Yajo:
The reasoning behind having several versions is that all of them have different `requirements.txt` files. In fact, you can see that the only real extra layer is the one that calls `install.sh`, which is the one that handles those differences. I guess they should be able to work by using different versions of the same requirements, but this way we have a place where to put the essential differences among versions.

This will become a little bit harder with v11+ that use python3, but we'll see that in the future.

We closed #38 previously because it was not backwards-compatible. We have tests today, you can PR again with tests and it should be merged 😉

lasley:
> We closed #38 previously because it was not backwards-compatible. We have tests today, you can PR again with tests and it should be merged 😉

I honestly don't have time to resubmit quite yet - it was completely backwards compatible upon first submission, but was refactored away from that upon request. The code was then squashed, and all the work on the backwards compatibility lost in the squash. I plan on redoing, but my fork is working just fine at the moment TBH.

Issue #58, opened 2017-07-26
Adding permissions for repo directories


lasley:
Sometimes Odoo needs write access to certain repo directories, such as in the instance of the static directory in Runbot.

One would assume that by creating a script in `build.d` with a numeric ordering after the repo aggregation in order to set the perms, it would be run after that. 

It seems however that the whole [custom build directory is run before the one included here](https://github.com/Tecnativa/docker-odoo-base/blob/master/bin/direxec.sh#L8), so the permission editing attempt takes place before the repos are aggregated into the directory. Is there a reasoning for this, or am I simply doing something wrong?

lasley:
Closing for PR #59

PR #59, opened 2017-07-26
[FIX] Run common directory before custom

lasley:
I originally submitted this as a question in #58, but in hindsight it should have been a PR.

Unless there is a specific reasoning for running the custom directory before common, I think we should reverse these. It stands to reason that custom logic should be able to override built-in, which is not the case here currently.


Yajo:
This has a reason: since the image has a lot of `ONBUILD` triggers, that get run just after the `FROM` clause always, we needed a way to add some customizations before anything else happened. That's why we have the `custom/build.d` and that's why we need it before anything else.

Besides, in a `custom/build.d` script you can remove a file from `common/build.d` if you need it.

If you still need further customizations to happen after the triggers, you can append those to your Dockerfile as usual:

```Dockerfile
FROM tecnativa/odoo-base:10.0
USER root
RUN chown -R odoo:odoo /opt/odoo
USER odoo
```

Another option is to add `ONBUILD ARG ODOO_OWNER=root:odoo` to be used in the `chown` parts, so you can override that in your builds with `odoo:odoo` to get what you desire.

All that said, do we still need this?

BTW, Travis is ❌

lasley:
Hmmmm that makes sense, thanks for the explanation. I wonder if it makes sense to maybe have a `before-build.d` and an `after-build.d` or something? I've run against this a few times, and while changing to root in the Dockerfile is an option - it sets off my linter.

I thought about some sort of file similar to the dependency ones that would allow you to list out directories that needed to be owned by Odoo, but that seemed like it was getting a bit into the over engineered side.

It should also be noted that I do not want all of the Odoo install owned by Odoo, only the Runbot static directory - that's where all the builds and stuff are stored. I think you have done the permissions well otherwise, Odoo really shouldn't be able to write to anywhere other than its filestore without being granted explicit permission - particularly because you then need to mount that folder as a volume in order to preserve the changes Odoo makes.

Yajo:
OK, then we can get to a better solution: sort alphabetically all `build.d/*` files, no matter if they're under `common` or `custom`. This way if you want to run something before, after, or in between some `common` scripts, you will just need to put the right order. We'd have to make our own `run-parts` implementation I guess, although the good thing is that [it could allow file extensions](https://github.com/Tecnativa/docker-odoo-base/blob/50ac4c8362aaa88adf260f5f9c2871547623c38a/README.md#why-my-99-whateversh-script-in-optodood-does-not-execute).

What do you think?

lasley:
I think that your's is the best solution & it's honestly how I thought it worked in the first place. As you know, I suck at writing ReadMes. You may not know, however, that I'm probably worse at reading them 😉 

Ima hit up good ol' Google and see if someone else has made a run-parts implementation meeting our spec. I bet we're not the first needing this. Otherwise, shouldn't be too difficult to write.

lasley:
On second thought, writing it from scratch was easier than Google 🚀

lasley:
This one's gtg now too

bin/direxec.sh
```diff
@@ -1,21 +0,0 @@
...
-        log INFO Executing contents of $dir
-        run-parts --exit-on-error $dir
-    else
-        log WARNING Skipping execution of $dir: not found
-    fi
-done
-
-if [ -n "$1" ]; then
-    set -x
-    exec "$@"
```
Yajo: I'm still missing this part. The entrypoint is supposed to execute any extra arguments as a command after doing its pre-tasks. Also this `exec "$@"` part makes the subprocess here take the PID of the parent process (which should be 1 in normal run contexts), making all signals work as expected with no extra work. Not sure how to mimic that in Python though, but we need it for sure.

Dockerfile
```diff
@@ -9,7 +9,7 @@ VOLUME ["/var/lib/odoo"]
 EXPOSE 8069 8072
 
 # Subimage triggers
-ONBUILD ENTRYPOINT ["/opt/odoo/common/entrypoint.sh"]
+ONBUILD ENTRYPOINT ["/opt/odoo/common/entrypoint"]
```
Yajo: Please call it `entrypoint.py`. Having to avoid file extensions with current `run-parts` implementation sucks. This also leads to removing [the corresponding FAQ](https://github.com/Tecnativa/docker-odoo-base#why-my-99-whateversh-script-in-optodood-does-not-execute).

bin/direxec
```diff
@@ -0,0 +1,27 @@
...
+    files_custom = [
+        os.path.join(dir_custom, d) for d in os.listdir(dir_custom)
+    ]
+
+files_all = files_common + files_custom
+sorted(files_all)
+
+for command in files_all:
+    if os.access(command, os.X_OK):
+        subprocess.check_output(command, shell=True)
```
lasley: I tried that originally & it was failing - I forget on which script now. Seems like some bash scripts require a shell? 🤷‍♂️

bin/direxec.sh
lasley: > makes the subprocess here take the PID of the parent process 
Hmmm so subprocess `shell=False` does this, but not `shell=True`. Looking at the `os.system` docs it explicitly says that it is running in a subshell, which I think would mean the same PID. I'm not exactly sure how to test though...

Dockerfile
lasley: Hmmmm why add a file extension? I used no extension here just in case we decide to switch the scripting language again for some reason. 

I don't feel like the filename of the binary should be intrinsically linked to the code type - the extension is a humanization factor that IMO is not really relevant when the pieces are managed with automation.

bin/direxec.sh
Yajo: Actually none of these do what you expect. I ran this inside a python container, see:

```
/ # python -c 'import os;os.system("sleep 5")' &
/ # ps aux
PID   USER     TIME   COMMAND
    1 root       0:00 sh
   16 root       0:00 python -c import os;os.system("sleep 5")
   17 root       0:00 sh -c sleep 5
   18 root       0:00 sleep 5
   19 root       0:00 ps aux
/ # python -c 'import subprocess;subprocess.check_output(["sleep","5"])' &
[1]-  Done                       python -c "import os;os.system(\"sleep 5\")"
/ # ps aux
PID   USER     TIME   COMMAND
    1 root       0:00 sh
   20 root       0:00 python -c import subprocess;subprocess.check_output(["sleep","5"])
   21 root       0:00 sleep 5
   22 root       0:00 ps aux
```

As you can see, `os.system` spawns the process inside a shell, and `subprocess.check_call` spawns it directly, but both spawn the subprocess. Check this instead:

```
/ # sh -c 'sleep 5' &
/ # ps aux
PID   USER     TIME   COMMAND
    1 root       0:00 sh
   11 root       0:00 sh -c sleep 5
   12 root       0:00 sleep 5
   13 root       0:00 ps aux
/ # sh -c 'exec sleep 5' &
[1]-  Done                       sh -c "sleep 5"
/ # ps aux
PID   USER     TIME   COMMAND
    1 root       0:00 sh
   14 root       0:00 sleep 5
   15 root       0:00 ps aux
```

By using `exec` in bash, you replace current process with the new one. The new one gets the parent's PID. Inside Docker, stopping a PID stops the container, that's why this little feature is important, so every signal you send with `docker kill` or with system's kill gets directly into the proper PID.

But I found [the solution](https://stackoverflow.com/a/6743663/1468388): 

```
/ # python -c 'import os; os.execvp("sleep", ["sleep", "5"])' &
/ # ps aux
PID   USER     TIME   COMMAND
    1 root       0:00 sh
   30 root       0:00 sleep 5
   31 root       0:00 ps aux
```

bin/direxec
```diff
@@ -0,0 +1,37 @@
...
+            raise EnvironmentError(
+                'Execution of "%s" failed with code "%d"' % (
+                    command, exit_code,
+                )
+            )
+
+extra_command = sys.argv[1:]
+
+if extra_command:
+    subprocess.check_output(extra_command)
```
Yajo: Replace with:

```python
try:
    os.execvp(sys.argv[1], sys.argv[1:])
except IndexError:
    pass
```

See explanation below.

Dockerfile
Yajo: OK, leave this one without extension, but please add it to scripts under `common/build.d`. It's not just humanization, it helps text editors too.

bin/direxec
```diff
@@ -0,0 +1,37 @@
...
+        os.path.join(dir_custom, d) for d in os.listdir(dir_custom)
+    ]
+except OSError:
+    pass
+
+for command in sorted(files):
+    if os.access(command, os.X_OK):
+        exit_code = os.system(command)
+        if exit_code:
+            raise EnvironmentError(
```
lasley: To clarify - the point of this instead of subprocess was to hopefully run under the same process ID. Given the below information, I will use `os.execvp` here instead of the recommended subprocess. 

The intent is to run all of the commands under the same pid right? I'm honestly thinking I'm still missing a nuance of why, so I'm just going with the flow. Is this so that Docker properly issues the exit command to the commands, which would allow for a proper shutdown instead of a hard one?

lasley: Ohhhhhh wait ok reading the docs of `execvp` & then your comments again - I totally get it! Thanks for the explanations

bin/direxec.py
```diff
@@ -0,0 +1,31 @@
...
+    pass
+
+for command in sorted(files):
+    if os.access(command, os.X_OK):
+        subprocess.check_call(command)
+
+extra_command = sys.argv[1:]
+
+if extra_command:
+    os.execvp('odoo', extra_command)
```
Yajo: Change `'odoo'` and put `extra_command[0]`.

We can also add a little smartypants feature here and do:

```python
if extra_command[0] == 'shell' or extra_command[0].startswith("-"):
    extra_command.insert(0, "odoo")
```

That would let us run commands like `docker-compose run --rm odoo shell` or `docker-compose run --rm odoo --version` instead of `docker-compose run --rm odoo odoo shell` and `docker-compose run --rm odoo odoo --version` (saving 1 `odoo`). Not a requirement though.

lasley:
> Still missing file extensions for common/build.d/* scripts, please add 😊

I really don't like this. Executables shouldn't have an extension IMO. Google seems to agree with me here, most of the top answers for "should executable scripts have an extension" explicitly say that it is bad practice.

Dockerfile
```diff
@@ -129,8 +129,8 @@ COPY build.d common/build.d
 COPY conf.d common/conf.d
 COPY entrypoint.d common/entrypoint.d
 RUN mkdir -p auto/addons custom/src/private \
-    && ln /usr/local/bin/direxec.sh common/entrypoint.sh \
-    && ln /usr/local/bin/direxec.sh common/build.sh \
+    && ln /usr/local/bin/direxec.py common/entrypoint \
+    && ln /usr/local/bin/direxec.py common/build \
```
Yajo: OK, then remove the extension here.

I agree executables under your `$PATH` should have no extension because you should not care about its language. About those under `build.d`, I think those are just scripts whose suffix should not matter much because they should not be executed direcly by the user, but OK, leave them as you wish 😉

bin/direxec.py
```diff
@@ -0,0 +1,33 @@
...
+for command in sorted(files):
+    if os.access(command, os.X_OK):
+        subprocess.check_call(command)
+
+extra_command = sys.argv[1:]
+
+if extra_command:
+    if extra_command[0] == 'shell' or extra_command[0].startswith("-"):
+        extra_command.insert(0, "odoo")
+    os.execvp(extra_command[0], extra_command)
```
Yajo: Still missing a test for this. Adding a `("--version",),` command would be enough.

Yajo:
I took the freedom to amend your commit to squash it with the suggestions I made, and added a test. I hope you don't mind. This one should fix the master branch too.

Dockerfile
lasley: For me it's a matter of how things look in Git. When I'm reviewing things, I pay attention to whether something is checked in with executable permissions, because they basically should never. No extension to me tells me that the file is known to be an executable and should have the permissions as such. An extension tells me there shouldn't be an executable bit. Maybe it's an oversimplification, but this is my main reason for pushing back.

PR #60, opened 2017-07-27
[IMP] Move config generation to entrypoint

### LINKED ISSUE #37: [RFC] Move config generation to entrypoint

lasley:
Turns out I never made any of the requested changes to remove the backwards compatibility in my forked image (hooray Git and laziness!). Resubmitting #37 in its original form, which is backwards compatible.

I'll still need to add a test for generating it on entrypoint. Should I just choose one of the scaffolds randomly, or do you have some sort of pattern you're going for with them in terms of what they test?

Fix #37

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ ONBUILD ARG DEPTH_DEFAULT=1
 ONBUILD ARG DEPTH_MERGE=100
 ONBUILD ARG CLEAN=true
 ONBUILD ARG COMPILE=true
-ONBUILD ARG CONFIG_BUILD=false
+ONBUILD ARG CONFIG_BUILD=true
 ONBUILD ARG PIP_INSTALL_ODOO=true
 ONBUILD ARG ADMIN_PASSWORD=admin
 ONBUILD ARG SMTP_SERVER=smtp
@@ -115,15 +115,15 @@ WORKDIR /opt/odoo
 RUN pip install --no-cache-dir \
     astor git-aggregator openupgradelib ptvsd==3.0.0 pudb wdb
 COPY bin/* /usr/local/bin/
-COPY bin/direxec.sh common/entrypoint.sh
-RUN ln common/entrypoint.sh common/build.sh
+RUN ln /usr/local/bin/direxec.sh common/build.sh
 COPY lib/odoobaselib /usr/local/lib/python2.7/dist-packages/odoobaselib
 COPY build.d common/build.d
 COPY conf.d common/conf.d
 COPY entrypoint.d common/entrypoint.d
 RUN mkdir -p auto/addons custom/src/private
 RUN chmod -R a+rx common/entrypoint* common/build* /usr/local/bin \
-    && chmod -R a+rX /usr/local/lib/python2.7/dist-packages/odoobaselib
+    && chmod -R a+rX /usr/local/lib/python2.7/dist-packages/odoobaselib \
+    && ln /usr/local/bin/direxec.sh common/entrypoint.sh
 
 # Execute installation script by Odoo version
 # This is at the end to benefit from cache at build time
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,10 +28,16 @@ ONBUILD ARG PGUSER=odoo
 ONBUILD ARG PGPASSWORD=odoopassword
 ONBUILD ARG PGHOST=db
 ONBUILD ARG PGDATABASE=prod
-ONBUILD ENV PGUSER="$PGUSER" \
+# Config variables
+ONBUILD ENV ADMIN_PASSWORD="$ADMIN_PASSWORD" \
+            UNACCENT="$UNACCENT" \
+            PGUSER="$PGUSER" \
             PGPASSWORD="$PGPASSWORD" \
             PGHOST="$PGHOST" \
-            PGDATABASE="$PGDATABASE"
+            PGDATABASE="$PGDATABASE" \
+            PROXY_MODE="$PROXY_MODE" \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -16,6 +16,7 @@ ONBUILD ARG DEPTH_DEFAULT=1
 ONBUILD ARG DEPTH_MERGE=100
 ONBUILD ARG CLEAN=true
 ONBUILD ARG COMPILE=true
+ONBUILD ARG CONFIG_BUILD=false
```
Yajo: I guess this should default to `true` to be backwards-compatible.

Dockerfile
```diff
@@ -113,7 +114,7 @@ RUN curl -SLo wkhtmltox.tar.xz https://github.com/wkhtmltopdf/wkhtmltopdf/releas
 WORKDIR /opt/odoo
 RUN pip install --no-cache-dir \
     astor git-aggregator openupgradelib ptvsd==3.0.0 pudb wdb
-COPY bin/autoaggregate bin/install.sh bin/log bin/pot bin/python-odoo-shell bin/unittest /usr/local/bin/
+COPY bin/* /usr/local/bin/
 COPY bin/direxec.sh common/entrypoint.sh
```
Yajo: If adding `bin/*`, then make here a `ln /usr/local/bin/direxec.sh common/entrypoint.sh`

Dockerfile
```diff
@@ -113,7 +114,7 @@ RUN curl -SLo wkhtmltox.tar.xz https://github.com/wkhtmltopdf/wkhtmltopdf/releas
...
 RUN ln common/entrypoint.sh common/build.sh
```
Yajo: ... and change here `common/entrypoint.sh` for `/usr/local/bin/direxec.sh` to make it clearer too.

entrypoint.d/50-config-generate
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [ "${CONFIG_BUILD}" = true ]; then
```
Yajo: This variable is a build arg, not an env. It means it will not be available at run time, so you better `test -f $OPENERP_SERVER`

entrypoint.d/50-config-generate
```diff
@@ -0,0 +1,8 @@
...
+    log WARNING Skipping Odoo config generation - performed at build time.
```
Yajo: ...  and then log `File $OPENERP_SERVER exists, skipping config generation`

Yajo:
BTW, make sure all the env vars used in `common/build.d` exist as `ENV` and not only as `ARG`, or it will not work neither.

entrypoint.d/50-config-generate
lasley: Hahaha interestingly this same recommendation was made and then reversed in the last PR. In the instance of config generation during entrypoint, we actually want to wipe any pre-existing configuration that may exist.

That said & in hindsight - I don't know of a way to change environment variables without destroying the container. A new container wouldn't have the configuration on it because it'd be a fresh image, so I think this means that we can actually use this method safely?

lasley:
This one's gtg

tests/scaffoldings/dotd/docker-compose.yaml
```diff
@@ -4,12 +4,18 @@ services:
...
+        environment:
+          PGUSER: another_odoo
+          PGPASSWORD: anotherodoopassword
+          PGHOST: postgresql
         depends_on:
-            - db
+            - postgresql
 
-    db:
+    postgresql:
```
Yajo: Leave the `db` name as it was please

Dockerfile
```diff
@@ -27,10 +28,16 @@ ONBUILD ARG PGUSER=odoo
...
+ONBUILD ENV ADMIN_PASSWORD="$ADMIN_PASSWORD" \
+            UNACCENT="$UNACCENT" \
+            PGUSER="$PGUSER" \
             PGPASSWORD="$PGPASSWORD" \
             PGHOST="$PGHOST" \
-            PGDATABASE="$PGDATABASE"
+            PGDATABASE="$PGDATABASE" \
+            PROXY_MODE="$PROXY_MODE" \
+            SMTP_SERVER="$SMTP_SERVER" \
+            WITHOUT_DEMO="$WITHOUT_DEMO"
```
Yajo: I'd appreciate a test that checks these values land correctly in the config file, both with args and envs 😊

Yajo: OK forget it, at the end of the day we only care about the config being present at the correct place at run time, checking the postgres connection works as you did is enough proof of that.

tests/scaffoldings/dotd/docker-compose.yaml
Yajo: Oh I get it! It's to check the default got overriden! Sorry 😅

PR #61, opened 2017-07-28
Added workers for production environment to enable longpolling

PCatinean:
First off a huge thank you for sharing this jewel. I've started using it on everything and life has become at least 5 times easier since.

As I don't have your expertise with Docker my PR's would be more on the line of humble suggestions than anything else.

Regarding the production environment it is mentioned in the doc that it needs quite a bit of customization but I thought that adding a minimum amount of workers to enable longpolling would be great.

I was not aware if I had a nginx configuration issue or something else before realizing that I did not deploy workers as longpolling was not working.

If the command is in the production environment the user can easily increase/decrease workers or add other extra commands.

I'm aware you can also place this in the configuration file but it's up to you to decide if it's worthwhile.

### DIFF
--- a/prod.yaml
+++ b/prod.yaml
@@ -13,6 +13,9 @@ services:
             - db
             - smtp
             - inbox
+        command:
+            - odoo
+            - --workers=2
         networks:
             default:
             inverseproxy_shared:


pedrobaeza:
Why don't you put the workers on the config file?

pedrobaeza:
OK, I didn't see your last paragraph...

PCatinean:
I only now noticed that test.yaml adds workers with the command as well, so I guess the design would be "workers in the conf are intended for production" and all other environments override this with command.

It's not a problem you can close the PR if you don't think it helps

PCatinean:
Also would a PR changing the backup service from docker-server-backup to docker-duplicity be of interest to you?

Yajo:
> I only now noticed that test.yaml adds workers with the command as well, so I guess the design would be "workers in the conf are intended for production" and all other environments override this with command.

That's exactly the idea 😉 

> Also would a PR changing the backup service from docker-server-backup to docker-duplicity be of interest to you?

I'm going to do it anyway, I just did not have time to test the duplicity version hard enough. If you are patient, the day will come for you 😆 

Thanks, pal! Since you auto-answered yourself, closing this one.

I'm glad you like the tool BTW, have a nice day!

PR #62, opened 2017-07-28
[IMP] Add some more default configurations

lasley:
* Add some more configurations, use the defaults from https://github.com/odoo/odoo/blob/3ace806d399e9e3abc5ada84ba98d2de100388d3/odoo/tools/config.py#L277

This is dependent on:
- [x] #60


pedrobaeza:
We should have both: the possibility to define these environment variables (not for us, but for @lasley that needs them), and default sane values.

Please rebase this PR and add these defaults.

Dockerfile
```diff
@@ -27,10 +28,38 @@ ONBUILD ARG PGUSER=odoo
...
+ONBUILD ARG ODOO_WORKERS=0
+ONBUILD ARG ODOO_LIMIT_TIME_REAL=120
+ONBUILD ARG ODOO_LIMIT_TIME_CPU=60
+ONBUILD ARG ODOO_LIMIT_MEMORY_HARD=805306368
+ONBUILD ARG ODOO_LIMIT_MEMORY_SOFT=671088640
+ONBUILD ARG ODOO_LIMIT_REQUEST=8192
+ONBUILD ARG ODOO_LIMIT_TIME_REAL_CRON=1
+ONBUILD ARG ODOO_DB_FILTER=
+ONBUILD ARG ODOO_LIST_DB=1
+ONBUILD ARG ODOO_LOG_LEVEL=info
```
Yajo: @pedrobaeza You can see the defaults he's adding here.

The problem is that for performance there are no *sane* defaults for everyone. If you apply a 32 core's sane defaults to a 4 core machine, it would get exhausted.

My original idea was to have some sane disabled defaults in config files that you could symlink in your `custom/conf.d` directory.

Imagine something like `/opt/odoo/common/disabled-conf.d/` with files like:

- 1vcpu.conf
- 2vcpu.conf
- 4vcpu.conf
- 8vcpu.conf (etc.)
- 4gbram.conf
- 8gbram.conf
- 32gbram.conf (etc.)

Similar to apache's `sites-available` vs `sites-enabled`.

So you can add those with `ln -s /opt/odoo/common/disabled-conf.d/{8vcpu,32gbram}.conf custom/conf.d` and have a community-maintained sane default for your machine. Or we could use an env var for that if it seems more convenient.

Dockerfile
```diff
@@ -27,10 +28,38 @@ ONBUILD ARG PGUSER=odoo
...
 ONBUILD ARG PGDATABASE=prod
-ONBUILD ENV PGUSER="$PGUSER" \
+ONBUILD ARG ODOO_MAX_CRON_THREADS=1
+ONBUILD ARG ODOO_WORKERS=0
+ONBUILD ARG ODOO_LIMIT_TIME_REAL=120
+ONBUILD ARG ODOO_LIMIT_TIME_CPU=60
+ONBUILD ARG ODOO_LIMIT_MEMORY_HARD=805306368
+ONBUILD ARG ODOO_LIMIT_MEMORY_SOFT=671088640
+ONBUILD ARG ODOO_LIMIT_REQUEST=8192
+ONBUILD ARG ODOO_LIMIT_TIME_REAL_CRON=1
```
Yajo: 1 second for crons? not much sane...

Dockerfile
```diff
@@ -27,10 +28,38 @@ ONBUILD ARG PGUSER=odoo
 ONBUILD ARG PGPASSWORD=odoopassword
 ONBUILD ARG PGHOST=db
 ONBUILD ARG PGDATABASE=prod
-ONBUILD ENV PGUSER="$PGUSER" \
+ONBUILD ARG ODOO_MAX_CRON_THREADS=1
+ONBUILD ARG ODOO_WORKERS=0
```
Yajo: Minimum 2 workers are required for production. Defaults should go for production always.

lasley: I explicitly aligned with the defaults of Odoo so as to not change the operation of anything from default. I feel that this is better than attempting to make some determination of what someone wants for this image.

Dockerfile
lasley: Oops this was `-1` in the [defaults](https://github.com/odoo/odoo/blob/3ace806d399e9e3abc5ada84ba98d2de100388d3/odoo/tools/config.py#L277). Guess I can't read!

Dockerfile
lasley: IMO because the variety of uses for this image, we shouldn't make too many assumptions about the deploy by default. 

Basically all I did here was expose some more configuration options, using specifically the ones [Odoo sets as default](https://github.com/odoo/odoo/blob/3ace806d399e9e3abc5ada84ba98d2de100388d3/odoo/tools/config.py#L277). 

I think the better configuration though would just be a part of the env files in scaffolds though:

```
odoo:
  env_file:
    - common.env
    - 32gbram.env
```

lasley: Note the above obviously wouldn't work if the configuration is being generated during build, but this also highlights why it should be in the endpoint instead. It seems pretty weird to statically define infrastructure capacity in a Docker image IMO

lasley:
Hmmm so the build is failing, but I don't see how this is something I could have caused with this change & I have absolutely no idea how to fix it. Ideas?

```
run-parts: failed to exec /opt/odoo/common/build.d/999-dependencies-cleanup: Text file busy
run-parts: /opt/odoo/common/build.d/999-dependencies-cleanup exited with return code 1
```

lasley:
(note I rebuilt three or four times under the hope it was a random error that would pass. no luck)

Yajo:
Please rebase

lasley:
Rebase looks to have cleared it

Dockerfile
Yajo: Still missing...

Dockerfile
```diff
@@ -28,6 +28,17 @@ ONBUILD ARG PGUSER=odoo
 ONBUILD ARG PGPASSWORD=odoopassword
 ONBUILD ARG PGHOST=db
 ONBUILD ARG PGDATABASE=prod
+ONBUILD ARG ODOO_MAX_CRON_THREADS=-1
```
Yajo: Default to 1.

Dockerfile
```diff
@@ -28,6 +28,17 @@ ONBUILD ARG PGUSER=odoo
...
+ONBUILD ARG ODOO_WORKERS=0
```
Yajo: Default to 2 (minimal required workers to get a functional worker-enabled odoo)

Dockerfile
```diff
@@ -28,6 +28,17 @@ ONBUILD ARG PGUSER=odoo
...
+ONBUILD ARG ODOO_LIMIT_TIME_REAL=120
+ONBUILD ARG ODOO_LIMIT_TIME_CPU=60
```
Yajo: In our experience it's quite better to increase these 2 defaults. Simple standard operations such as downloading a db backup never get done with these values. At least 10x please.

conf.d/60-performance.conf
```diff
@@ -0,0 +1,9 @@
+[options]
+max_cron_threads=$ODOO_MAX_CRON_THREADS
+workers=$ODOO_WORKERS
+limit_time_real=$ODOO_LIMIT_TIME_REAL
+limit_time_cpu=$ODOO_LIMIT_TIME_CPU
+limit_memory_hard=$ODOO_LIMIT_MEMORY_HARD
+limit_memory_soft=$ODOO_LIMIT_MEMORY_SOFT
+limit_request=$ODOO_LIMIT_REQUEST
+limit_time_real_cron=$ODOO_LIMIT_TIME_REAL_CRON
```
Yajo: Please surround `=` signs with spaces like in the other config files.

conf.d/70-security.conf
```diff
@@ -0,0 +1,3 @@
+[options]
+dbfilter=$ODOO_DB_FILTER
+list_db=$ODOO_LIST_DB
```
Yajo: spaces here too.

BTW, if you are naming this file "security", please move `admin_passwd` here, to be consistent, please.

conf.d/80-logging.conf
```diff
@@ -0,0 +1,2 @@
+[options]
+log_level=$ODOO_LOG_LEVEL
```
Yajo: space

lasley:
> I'm suggesting some little changes on the defaults. They diverge from Odoo's, but seem more sensible under our experience. Maybe @pedrobaeza wants to add some further suggestions before merging?

I really don't like changing the Odoo defaults. Why should we make the users memorize a completely new set of defaults? I agree that the Odoo ones aren't sensible in some scenarios, but that is what they are & they don't actually hurt anything. 

IMO the scaffolds are where we should be making configuration decisions. We just need a way to expose them to the user so that they're not remaking all of these scripts.

Furthermore, I don't even agree with one of the recommended default changes for the worker time. Our deploys use isolated cron and service instances that allow for long running operations. Web facing instances should not allow this type of thing, which can be an avenue for DOS.

lasley:
Regarding the duplications in configs - the last configuration item defined takes precedence.

Which order do the configs drop in? Common then custom? or is it Custom then Common like it was in build/entry?

Yajo:
> Regarding the duplications in configs - the last configuration item defined takes precedence.

Can we have a test on that please? You can test it with a simple odoo shell script.

> Which order do the configs drop in?

[It's common and then custom](https://github.com/Tecnativa/docker-odoo-base/blob/master/bin/config-generate#L4).

lasley:
Ok so I think that this works out great then with the custom coming after common, it can be completely controlled externally.

Will add test, although I'm not quite sure what you mean by a shell script? I'll submit what I'm thinking, but I don't think it aligns with what you are because there's no script.

tests/scaffoldings/dotd/docker-compose.yaml
```diff
@@ -10,9 +10,8 @@ services:
     environment:
       PGUSER: another_odoo
       PGPASSWORD: anotherodoopassword
-      PGHOST: postgresql
     depends_on:
-      - postgresql
+      - postgresql:manual_db_host
```
Yajo: BTW [This is not a valid syntax](https://travis-ci.org/Tecnativa/docker-odoo-base/builds/262422829#L4924-L4934), although you have to revert this anyway (see above)

tests/scaffoldings/dotd/docker-compose.yaml
```diff
@@ -10,9 +10,8 @@ services:
     environment:
       PGUSER: another_odoo
       PGPASSWORD: anotherodoopassword
-      PGHOST: postgresql
```
Yajo: I think some entrypoint script will fail if you set an invalid `$PGHOST` anyway.

I think I will explain better the test I expected:

1. Add to `environment`: 
   ```yaml
   ODOO_LIMIT_REQUEST: 8000
   ODOO_LIST_DB: 0
   ```
1. Add a `dotd/custom/conf.d/request_limit.conf` with:
   ```
   [options]
   limit_request = 9000
   ```
1. Add `dotd/custom/entrypoint.d/600-goodconf.py`:

   ```python
   #!/usr/local/bin/python-odoo-shell
   from odoo.tools import config
   assert config.get("limit_request") == 9000
   assert not config.get("list_db")
   ```

This way we make sure the env variables get into config, and that manual configs override them.

lasley:
It seems Travis keeps failing on the below & it seems to be across all my PRs. Any ideas?

```
WARNING:docker-odoo-base:Blindly symlinking odoo executable
  File "/opt/odoo/common/build", line 26, in <module>
  File "/usr/lib/python2.7/subprocess.py", line 535, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/usr/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/usr/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 26] Text file busy
```

Yajo:
Can you rebase please?

lasley:
Hmmm so do we have any idea what's causing the text file error? Should I just rebase until it passes? It totally seems like something intermittent that is spread across all PRs

Yajo:
See #65, which I'll reopen now.

lasley:
Is there anything else I need to do here other than rebase?

conf.d/60-performance.conf
```diff
@@ -0,0 +1,9 @@
+[options]
```
Yajo: You have to remove this here and in the other files. It fails in Odoo 11

Dockerfile
```diff
@@ -29,6 +29,17 @@ ONBUILD ARG PGUSER=odoo
 ONBUILD ARG PGPASSWORD=odoopassword
 ONBUILD ARG PGHOST=db
 ONBUILD ARG PGDATABASE=prod
+ONBUILD ARG ODOO_MAX_CRON_THREADS=2
+ONBUILD ARG ODOO_WORKERS=0
```
Yajo: Since the image should be _production first_, I guess we can put 4 here, or at least 2 (minimum required for odoo to work and print reports).

Dockerfile
```diff
@@ -29,6 +29,17 @@ ONBUILD ARG PGUSER=odoo
...
+ONBUILD ARG ODOO_LIMIT_TIME_REAL=120
+ONBUILD ARG ODOO_LIMIT_TIME_CPU=60
+ONBUILD ARG ODOO_LIMIT_MEMORY_HARD=805306368
+ONBUILD ARG ODOO_LIMIT_MEMORY_SOFT=671088640
+ONBUILD ARG ODOO_LIMIT_REQUEST=8192
+ONBUILD ARG ODOO_LIMIT_TIME_REAL_CRON=-1
+ONBUILD ARG ODOO_DB_FILTER=""
+ONBUILD ARG ODOO_LIST_DB=1
+ONBUILD ARG ODOO_LOG_LEVEL=info
```
Yajo: Do you think we can reuse [image's LOG_LEVEL](https://github.com/Tecnativa/docker-odoo-base/pull/62/files#diff-3254677a7917c6c01f55212f86c57fbfR67)? Or do you prefer to keep them separate?

pedrobaeza:
Any news with this?

Dockerfile
lasley: IMO it would be best to keep them separate. The image log to me is a build/boot thing, vs the Odoo log being runtime. While they may be coordinated, they're for different people to use (the builder vs the maintainer)

lasley:
Updated. Sorry for the lag

8.0.Dockerfile
```diff
@@ -29,6 +29,17 @@ ONBUILD ARG PGUSER=odoo
 ONBUILD ARG PGPASSWORD=odoopassword
 ONBUILD ARG PGHOST=db
 ONBUILD ARG PGDATABASE=prod
+ONBUILD ARG ODOO_MAX_CRON_THREADS=2
+ONBUILD ARG ODOO_WORKERS=2
```
Yajo: I think 4 is a better default here. 2 should be the minimal amount of workers to get a prefork server up and running, but even servers with a single CPU should be able to handle 4 workers nowadays, and it seems a better production default. The devel and test environments override this default already:

https://github.com/Tecnativa/docker-odoo-base/blob/447f57c547896b2212f68db8fa5308da7473bd99/devel.yaml#L36

https://github.com/Tecnativa/docker-odoo-base/blob/447f57c547896b2212f68db8fa5308da7473bd99/test.yaml#L34

8.0.Dockerfile
```diff
@@ -29,6 +29,17 @@ ONBUILD ARG PGUSER=odoo
 ONBUILD ARG PGPASSWORD=odoopassword
 ONBUILD ARG PGHOST=db
 ONBUILD ARG PGDATABASE=prod
+ONBUILD ARG ODOO_MAX_CRON_THREADS=2
```
Yajo: I think a better default is 1. 2 crons can collide each other and produce more concurrency errors, and besides cron jobs are usually not that important that can't be delayed a little bit until the previous job finishes. It's better to default to 1 and let customizations increase that number.

8.0.Dockerfile
```diff
@@ -29,6 +29,17 @@ ONBUILD ARG PGUSER=odoo
...
+ONBUILD ARG ODOO_WORKERS=2
+ONBUILD ARG ODOO_LIMIT_TIME_REAL=120
+ONBUILD ARG ODOO_LIMIT_TIME_CPU=60
```
Yajo: With these limits it's quite common that you cannot download a database backup. I'd increase both to 5 minutes minimum by default.

tests/scaffoldings/dotd/custom/conf.d/request_limit.conf
```diff
@@ -0,0 +1,2 @@
+[options]
```
Yajo: Remove this line or it will fail in odoo v11

8.0.Dockerfile
lasley: IMO it's best to leave these at Odoo default and let a developer decide. I run my productions with these limits for security purposes. There is no reason a database backup should be able to complete from a production instance using user-facing methods

lasley:
> Do we really need to prefix all of these with ODOO_? I think that's implicit...

IIRC there were some collisions, so I thought it best for consistency purposes. There's a lot more than just Odoo variables in this env

lasley:
Done. /crosses_fingers

tests/scaffoldings/dotd/custom/entrypoint.d/500-create-db.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+# Need a preexisting DB to run odoo shell scripts
+psql -d postgres -c "CREATE DATABASE $PGDATABASE"
```
Yajo: Delete this file. See below.

tests/scaffoldings/dotd/custom/entrypoint.d/600-goodconf.py
```diff
@@ -0,0 +1,4 @@
+#!/usr/local/bin/python-odoo-shell
+from odoo.tools import config
```
Yajo: There's a problem on putting these test files at the entrypoint: both get executed before each test in `dotd`, and there are a lot.

Put this file in a separate script `dotd/custom/extras/goodconf.py`. Then [add a command to the list](https://github.com/Tecnativa/docker-odoo-base/blob/master/tests/__init__.py#L194) that executes this script. Make sure you add the command after another test that should have created a database, this way you don't have to worry about the DB creation thing.

tests/scaffoldings/dotd/custom/entrypoint.d/600-goodconf.py
```diff
@@ -0,0 +1,4 @@
...
+assert config.get("limit_request") == 9000
```
Yajo: I'm afraid we will need some more magic to make this test pass, because [on v11 you cannot have duplicate config entries](https://travis-ci.org/Tecnativa/docker-odoo-base/jobs/316530702#L3757), 😕 possibly a new limitation on Python 3 implementation.

Merging as is would break v11 production deployments, not nice really... (Blessed be tests!)

lasley: Well I think we just need another configuration variable. I've now exhausted all the ones that exist in Odoo, so I'll just make one up I think

lasley:
Well hmmmm- I put this command last in the stack in hopes that it would be after a DB creation, but that does not seem to be the case. I'm not sure I understand this testing architecture TBH

11.0.Dockerfile
```diff
@@ -27,6 +27,17 @@ ONBUILD ARG PGUSER=odoo
...
+ONBUILD ARG ODOO_WORKERS=4
+ONBUILD ARG ODOO_LIMIT_TIME_REAL=120
+ONBUILD ARG ODOO_LIMIT_TIME_CPU=60
+ONBUILD ARG ODOO_LIMIT_MEMORY_HARD=805306368
+ONBUILD ARG ODOO_LIMIT_MEMORY_SOFT=671088640
+ONBUILD ARG ODOO_LIMIT_REQUEST=8192
+ONBUILD ARG ODOO_LIMIT_TIME_REAL_CRON=-1
+ONBUILD ARG ODOO_DB_FILTER=""
+ONBUILD ARG ODOO_LIST_DB=1
+ONBUILD ARG ODOO_LOG_LEVEL=info
```
Yajo: You forgot to translate these `ARG` into `ENV`, like in the v8 dockerfile. That's [the cause of v11 failure](https://travis-ci.org/Tecnativa/docker-odoo-base/jobs/317022563#L4426).

tests/__init__.py
```diff
@@ -192,6 +192,8 @@ def test_dotd(self):
                 ("odoo", "--version"),
                 # Implicit ``odoo`` command also works
                 ("--version",),
+                # Environment vars were converted to config
+                ( "custom/extras/goodconf.py", ),
```
Yajo: It seems that `odoo shell` does not autocreate a database like normal odoo commands do from v9+. Then you can change this for `("bash", "-c", "odoo --stop-after-init && custom/extras/goodconf.py")` or something similar that creates the DB before executing the shell script.

Yajo:
Yikes! Well, now we're getting into it. 😉 

The `test_dotd()` test didn't have a single test that needed a present Odoo DB.

Now that we need it to create a Odoo db, we need to make the required steps for it. It seems we're hitting 2 problems:

1. [In v8, you still need to precreate the database](https://travis-ci.org/Tecnativa/docker-odoo-base/jobs/318356658#L5104). You don't have to do it in higher versions, so we need a way to run that step only in v8, and the testing suite provides such tool. You need to imitate what `test_smallest()` does: https://github.com/Tecnativa/docker-odoo-base/blob/318cf1a65c6977418299e7936d68fcde72c19ef3/tests/__init__.py#L147-L155
2. In v9+ the logs appear to indicate that [the `__manifest__.py` or `__openerp__.py` files are empty](https://travis-ci.org/Tecnativa/docker-odoo-base/jobs/318356660#L5131-L5140). You can choose between actually filling those files found in the `dotd` scaffolding, or add a new scaffolding folder to only test the config builder. Maybe the latter would be easier.

Yajo:
I'm not sure if problem 2 would be fixed with a rebase now...

lasley:
Hmmm so v8 is still failing due to no db, but I very clearly see the create database command success from Postgres. I'm taking a total stab in the dark, but I'm thinking there's a different PSQL container for each matrix iteration, so I just created the db in all versions instead 🤷‍♂️

tests/scaffoldings/dotd/custom/extras/goodconf.py
```diff
@@ -0,0 +1,4 @@
+#!/usr/local/bin/python-odoo-shell
+from odoo.tools import config
+assert config.get("test_config_setting") == "Over9000"
```
Yajo: [It says line 3 fails](https://travis-ci.org/Tecnativa/docker-odoo-base/jobs/327450411#L4695-L4696), so this is possibly not doing what you expect.

lasley: I'm actually not sure why I'm even creating that file or testing this part. Totally unrelated to the task at hand. Fixed.

If tests pass though, I think this might mean there's an issue with config generation in the custom directories.

tests/scaffoldings/dotd/custom/extras/goodconf.py
```diff
@@ -0,0 +1,4 @@
+#!/usr/local/bin/python-odoo-shell
+from odoo.tools import config
+assert config.get("limit_request") == "8000"
```
Yajo: This still [fails on line 3](https://travis-ci.org/Tecnativa/docker-odoo-base/jobs/327804781#L5232-L5233).

I'm suspecting that duplicate keys cannot be correctly parsed, or that the parameter is converted to `int` and thus this assertion fails.

Maybe could you add assertion messages that help us to debug further? Example:

```python
assert config.get("limit_request") == 8000, "%r is not %r" % (config.get("limit_request"), 8000)
```

... and the same in line 4

Yajo:
Hi friend, I added a scaffolding just to test settings in https://github.com/Tecnativa/docker-odoo-base/commit/52ad12652f0f6b64c8e813194c6e81bfd0a60996, and it always uses a database. It's only testing versions >= 9.0 because older did not have odoo shell.

Most likely you can take advantage of that to do the tests in an easier way 😊

8.0.Dockerfile
```diff
@@ -48,7 +59,19 @@ ONBUILD ENV ADMIN_PASSWORD="$ADMIN_PASSWORD" \
...
+            ODOO_LIMIT_TIME_REAL="$ODOO_LIMIT_TIME_REAL" \
+            ODOO_LIMIT_TIME_CPU="$ODOO_LIMIT_TIME_CPU" \
+            ODOO_LIMIT_MEMORY_HARD="$ODOO_LIMIT_MEMORY_HARD" \
+            ODOO_LIMIT_MEMORY_SOFT="$ODOO_LIMIT_MEMORY_SOFT" \
+            ODOO_LIMIT_REQUEST="$ODOO_LIMIT_REQUEST" \
+            ODOO_LIMIT_TIME_REAL_CRON="$ODOO_LIMIT_TIME_REAL_CRON" \
+            ODOO_DB_FILTER="$ODOO_DB_FILTER" \
+            ODOO_LIST_DB="$ODOO_LIST_DB" \
+            ODOO_LOG_LEVEL="$ODOO_LOG_LEVEL"
+>>>>>>> [IMP] Add some more default configurations
```
Yajo: oops.. wrong rebase

tests/__init__.py
```diff
@@ -140,6 +140,9 @@ def test_settings(self):
             ("--stop-after-init",),
             # SMTP settings work
             ("./custom/scripts/test_smtp_settings.py",),
+            # Environment vars were converted to config
+            ("bash", "-c",
+             "odoo --stop-after-init && custom/extras/goodconf.py"),
```
Yajo: The `--stop-after-init` is done above, just call `goodconf`

Yajo:
This is very outdated and stale. Closing.

PR #64, opened 2017-08-01
Add autoupdate utility

Yajo:
You'll ❤️ this @lasley

### DIFF
--- a/README.md
+++ b/README.md
@@ -704,6 +704,25 @@ Only Odoo's:
 
     docker-compose run --rm odoo odoo -u addon1,addon2 --stop-after-init
 
+##### Update changed addons only
+
+Add `module_auto_update` from https://github.com/OCA/server-tools to your
+installation following the standard methods of `repos.yaml` + `addons.yaml`.
+
+Now we will install the addon:
+
+    docker-compose up -d
+    docker-compose run --rm odoo --stop-after-init -u base
+    docker-compose run --rm odoo --stop-after-init -i module_auto_update
+    docker-compose restart odoo
+
+It will automatically update addons that got updated every night.
+To force that automatic update now in a separate container:
+
+    docker-compose up -d
+    docker-compose run --rm odoo autoupdate
+    docker-compose restart odoo
+
 ##### Export some addon's translations to stdout
 
     docker-compose run --rm odoo pot addon1[,addon2]
--- a/bin/autoupdate
+++ b/bin/autoupdate
@@ -1,0 +1,2 @@
+#!/usr/local/bin/python-odoo-shell
+env["base.module.upgrade"].upgrade_module()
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -704,6 +704,25 @@ Only Odoo's:
...
+    docker-compose run --rm odoo --stop-after-init -u base
+    docker-compose run --rm odoo --stop-after-init -i module_auto_update
+    docker-compose restart odoo
+
+It will automatically update addons that got updated every night.
+To force that automatic update now in a separate container:
+
+    docker-compose up -d
+    docker-compose run --rm odoo autoupdate
+    docker-compose restart odoo
```
pedrobaeza: This is not needed if you run previously `up -d`

Yajo: It is needed, because an updated addon could change models or force a cache clearing that could affect production instance.

Maybe in some cases it will not be needed, but since this is a recipe that should work "always", it's best here.

pedrobaeza: OK, I don't know any case where you need to restart if fresh Python code is already loaded, but anyway...

Yajo: It's very simple: i.e. an addon adds a new dependency, you update it and it gets installed, the registry gets reloaded, but only in memory, in the updating container, not in the previously running one.

pedrobaeza: No, that's not a valid case. In fact, in the case you describe would result in an error message saying that the module can't be loaded as there's a dependency missing, and it won't load nor update. You need to manually install the dependency. This is of course one problem for the autoupdater module: if you add new dependencies to modules, it won't be auto-installed.

Yajo: In v10 there's also the templates cache

pedrobaeza: OK, no need to continue debating about this. Let's keep this instruction for any possible case.

pedrobaeza:
Why travis is red?

README.md
lasley: I can elaborate on this problem in our production instances. In our case, we run 4 Odoo containers - web, website, cron, longpolling.

We're using a strategy very similar to this one at the moment, but originally just started with another container running the `--update=all` command. 

We have found that we need to restart all of the other Odoo instances after the update container stops, regardless of the Python code that is residing in the other Odoo containers. The main thing that seems to be screwed up is all of the XML caching and minimization.

lasley: It's also worth noting that now that our strategy does use the `module_auto_update`, we explicitly disable the cron. 

Might be worth mentioning that in the instructions - when we tested it, the database would sometimes deadlock (in the event of a lot of updates). The deadlocks would cause our health checks to restart the containers, which would then quadruple down on the amount of Odoo instances trying to update the same database. It would have been a real mess had it not been in UAT.

Yajo:
Not sure, it seems to fail randomly, I'll rebuild.

Yajo:
[The random failure is here](https://travis-ci.org/Tecnativa/docker-odoo-base/builds/259716971#L6975-L6976), I'm hoping that #59 fixes it with its new run-parts implementation, but honestly I don't know why it happens... Let's wait for #59 and see.

PR #68, opened 2017-08-08
Update backup and config generation systems

### LINKED ISSUE #63: Update standard scaffolding to use entrypoint config generation
After merging #60, we need to update scaffolding to make that one the default and deprecate old system.

Yajo:
- Use tecnativa/duplicity:postgres-s3 backup image by default.
- Use new entrypoint configuration generation system from #60.
- Fix #63.

### DIFF
--- a/.env
+++ b/.env
@@ -36,16 +36,9 @@ INBOX_REAL_LISTEN=:995
 INBOX_REAL_TALK=inbox.example.com:995
 
 # Backup
-BACKUP_ROOT_EMAIL_FROM=backup@example.com
-BACKUP_ROOT_EMAIL_TO=alerts@example.com
 BACKUP_S3_ACCESS_KEY=example key
 BACKUP_S3_SECRET_KEY=example secret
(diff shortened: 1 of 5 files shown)


lasley:
Any reason to not just use Camptocamp's solution, [Conplicity](https://github.com/camptocamp/conplicity)? We've been using it for a few months, and I find it quite awesome.

Yajo:
What does it offer above plain duplicity?

lasley:
> What does it offer above plain duplicity?

Just automation scripts wrapped around Duplicity. The big one is detection of the data types on the volume, with support for DB dumps from most of the standard database types. This in comparison to running the DB dump (mysql_dump, pg_dump, etc) and subsequently backing up the dump.

Yajo:
Well... maybe if our version wasn't already done... but it is, and it's much simpler IMHO.

In any case, it's only part of the scaffolding. You can use your own backup tool or scaffolding (see https://github.com/Tecnativa/docker-odoo-base#can-i-have-my-own-scaffolding).

PR #82, opened 2017-09-21
[DCK] No volume versions

Yajo:
This is a **breaking change**.

The purpose of this feature was that you were able to quickly switch branches in the project by using git, and avoid different branches to share the same volumes, since that can lead to errors in different Odoo versions.

At the end of the day, it turns out that feature was not used. Most version switching happens in devel, where you'd have to re-run `setup-devel` stuff, which slowed down the devel process too much.

Other version switches happened when running OpenUpgrade among versions, and in these cases we need to share the same volume among versions, so this "feature" was becoming a "bug".

So, finally we decided to drop it. This eases the initial scaffolding setup (less things to change). If we need different versions, it's as easy as having different clones, one on each branch, or switching the env `COMPOSE_PROJECT` variable.

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -13,7 +13,7 @@ services:
             PYTHONOPTIMIZE: 2
         tty: true
         volumes:
-            - filestore$ODOO_MAJOR:/var/lib/odoo:z
+            - filestore:/var/lib/odoo:z
         labels:
             traefik.docker.network: "inverseproxy_shared"
             traefik.enable: "true"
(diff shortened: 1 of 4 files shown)


Yajo:
Check out https://github.com/Tecnativa/docker-odoo-base/issues/67#issuecomment-331098437 to avoid retrocompatibility problems.

pedrobaeza:
Have you tried this? I tried in one project and there was an error because in some point it requires db numbered volume.

In Docker, isn't there the possibility of renaming existing volumes?

Yajo:
It should be fixed now.

> In Docker, isn't there the possibility of renaming existing volumes?

Not AFAIK

pedrobaeza:
This breaking change has been merged, so this should be commented for letting everyone knows about that (merge operations are not sent).

Yajo:
Yes, it was advertised at https://github.com/Tecnativa/docker-odoo-base/issues/67#issuecomment-331098437

PR #85, opened 2017-10-02
Support Odoo v11 🎉

### LINKED ISSUE #22: Dependencies for Pillow (Debian stretch)
I was having trouble installing Pillow using buildout. FYI, here I found libs that are required:
https://github.com/python-pillow/docker-images/blob/master/debian-stretch-x86/Dockerfile

apt-get -y \
    install xvfb sudo \
    git wget python-numpy python-scipy netpbm \
    python-qt4 ghostscript libffi-dev libjpeg-turbo-progs \
    python-setuptools python-virtualenv \
    python-dev python3-dev cmake \
    libtiff5-dev libjpeg62-turbo-dev zlib1g-dev \
    libfreetype6-dev liblcms2-dev libwebp-dev tcl8.6-dev tk8.6-dev \
    python-tk python3-tk

Yajo:
wip
closes #22


8.0.Dockerfile
```diff
@@ -0,0 +1 @@
+10.0.Dockerfile
```
lasley: IMO the 8.0 should be the one containing the Dockerfile, with the older ones inheriting. This would align with what we will do when the 12.0 comes out, inheriting the old unless there is some new installation.

11.0.Dockerfile
```diff
@@ -0,0 +1,152 @@
...
+# See https://github.com/$ODOO_SOURCE/blob/$ODOO_VERSION/debian/control
+RUN apt-get update \
+    && apt-get -y upgrade \
+    && apt-get install -y --no-install-recommends \
+        python3 ruby-compass \
+        fontconfig libfreetype6 libxml2 libxslt1.1 libjpeg62-turbo zlib1g \
+        libfreetype6 liblcms2-2 libtiff5 tk tcl libpq5 \
+        libldap-2.4-2 libsasl2-2 libx11-6 libxext6 libxrender1 \
+        locales-all zlibc \
+        bzip2 ca-certificates curl gettext-base git gnupg2 nano \
```
lasley: > gnupg2

Whoah. We gained something GPG in Odoo 11?!

Yajo: No, it's just required to install node + npm below. It seems debian 9 doesn't bundle npm.

lasley:
😆  I love the cascading fixups. They're almost entrancing

tests/__init__.py
```diff
@@ -194,6 +185,9 @@ def test_dotd(self):
                 ("--version",),
             )
 
+    @unittest.skipUnless(
```
lasley: TIL this decorator

PR #86, opened 2017-10-03
[WIP] Autogenerate repos.yaml for missing addons

PCatinean:
…r automatic dependency resolution


build.d/300-oca-dependencies
```diff
@@ -0,0 +1,89 @@
...
+    for line in depfile:
+        line = line.strip()
+        if not line or line.startswith('#'):
+            continue
+        parts = line.split()
+        repo = parts[0]
+        if len(parts) > 2:
+            branch = parts[2]
+        else:
+            branch = os.environ.get('VERSION', '10.0')
```
Yajo: This env var is called `ODOO_VERSION`, and you should not default to any value. It is considered an error if it is not available. Just use brackets and let the proper `KeyError` bubble up if it ever happens.

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,89 @@
...
+import yaml
+import glob
+import urllib2
+
+from odoobaselib import CUSTOM_DIR, ADDONS_DIR
+
+
+# Method from maintainer-quality-tools
+def parse_depfile(depfile, owner='OCA'):
+    deps = []
```
Yajo: I think this should be a set, to avoid possible duplicates.

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,89 @@
...
+            url = parts[1]
+        else:
+            url = 'https://github.com/%s/%s.git' % (owner, repo)
+        deps.append((repo, url, branch))
+    return deps
+
+
+def process_dependencies(dep_list):
+    for dep in dep_list:
+        if dep in processed_dependencies:
```
Yajo: declare this in the method?

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,89 @@
...
+        except Exception as e:
+            # Not neccesairly error but it might not have oca_dependencies.txt
+            print('Error fetching url %s. Error: %s' % (
+                oca_dep_url, e.code)
+            )
+            # TODO: Retry mechanism
+    return True
+
+addons_yaml = yaml.load(open('%s/addons.yaml' % CUSTOM_DIR, 'r'))
+repos_yaml = yaml.load(open('%s/repos.yaml' % CUSTOM_DIR, 'r'))
```
Yajo: Get these from odoobaselib

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,89 @@
...
+
+branch = os.environ.get('VERSION', '10.0')
```
Yajo: Same as above, no `.get`!

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,89 @@
...
+processed_dependencies = set()
+base_url = 'https://raw.githubusercontent.com/OCA'
```
Yajo: This should be a constant at the top  of the file

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,89 @@
...
+
+# Find all dependency files after initial aggregation
+dep_files = glob.glob('%s/**/oca_dependencies.txt' % CUSTOM_DIR)
```
Yajo: Use iglob

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,89 @@
...
+# Get all initial repositories across all projects
+for dep in dep_files:
+    file_obj = open(dep, 'r')
+    deps = set(parse_depfile(file_obj))
+    initial_dependencies.update(deps)
+
+process_dependencies(initial_dependencies)
+
+auto_addons_yaml = open('%s/addons.yaml' % ADDONS_DIR, 'w')
+auto_repos_yaml = open('%s/repos.yaml' % ADDONS_DIR, 'w')
```
Yajo: Declare these at odoobaselib

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,89 @@
...
+for dep in processed_dependencies:
+    addons_yaml[dep[0]] = ['*']
+
+    repo_vals = {
+        'defaults': {'depth': '$DEPTH_DEFAULT'},
+        'remotes': {'origin': dep[1]},
+        'target': {'origin': '$ODOO_VERSION'},
+        'merges': [{'origin': '$ODOO_VERSION'}],
+    }
+    repos_yaml[dep[0]] = {'./%s' % dep[0]: repo_vals}
```
Yajo: Remove the `./` please

PCatinean:
@Yajo added the mentioned changes. However in local tests it fails, i need to debug to fix them (one being it should not run if AGGREGATE env var is not defined). It also has to start the section aggregation once everything worked

docker-compose.yml
```diff
@@ -0,0 +1 @@
+devel.yaml
```
Yajo: Remove this file.

build.d/300-oca-dependencies
```diff
@@ -20,47 +31,45 @@ def parse_depfile(depfile, owner='OCA'):
...
-repos_yaml = yaml.load(open('%s/repos.yaml' % CUSTOM_DIR, 'r'))
-
-branch = os.environ.get('VERSION', '10.0')
-processed_dependencies = set()
-base_url = 'https://raw.githubusercontent.com/OCA'
+branch = os.environ['ODOO_VERSION']
 
 # Find all dependency files after initial aggregation
-dep_files = glob.glob('%s/**/oca_dependencies.txt' % CUSTOM_DIR)
+dep_files = iglob('%s/**/oca_dependencies.txt' % CUSTOM_DIR)
```
Yajo: No need for double asterisk, I think the file can only be found at repo root

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,97 @@
...
+        parts = line.split()
+        repo = parts[0]
+        if len(parts) > 2:
+            branch = parts[2]
+        else:
+            branch = os.environ['ODOO_VERSION']
+        if len(parts) > 1:
+            url = parts[1]
+        else:
+            url = 'https://github.com/%s/%s.git' % (owner, repo)
```
Yajo: You should add to the dockerfile `ONBUILD ARG DEFAULT_REPO_PATTERN="https://github.com/OCA/%s.git"` and rely on that env variable instead.

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,97 @@
...
+dependencies = process_dependencies(initial_dependencies)
+
+# Generate auto_addons and auto_repos yaml files
+for dep in dependencies:
+    ADDONS_YAML[dep[0]] = ['*']
+
+    repo_vals = {
+        'defaults': {'depth': '$DEPTH_DEFAULT'},
+        'remotes': {'origin': dep[1]},
+        'target': {'origin': '$ODOO_VERSION'},
```
Yajo: [The target is a string](https://github.com/acsone/git-aggregator#configuration-file).

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,97 @@
...
+    file_obj = open(dep, 'r')
+    deps = set(parse_depfile(file_obj))
+    initial_dependencies.update(deps)
+
+# Get all subsequent repositories from oca_dependencies via github
+dependencies = process_dependencies(initial_dependencies)
+
+# Generate auto_addons and auto_repos yaml files
+for dep in dependencies:
+    ADDONS_YAML[dep[0]] = ['*']
```
Yajo: `ADDONS_YAML` only contains the string of the file location. The actual parsing is done at `odoobaselib.addons_config()`

build.d/300-oca-dependencies
```diff
@@ -0,0 +1,97 @@
...
+
+    repo_vals = {
+        'defaults': {'depth': '$DEPTH_DEFAULT'},
+        'remotes': {'origin': dep[1]},
+        'target': {'origin': '$ODOO_VERSION'},
+        'merges': [{'origin': '$ODOO_VERSION'}],
+    }
+    REPOS_YAML[dep[0]] = {'%s' % dep[0]: repo_vals}
```
Yajo: `REPOS_YAML` contains only the file string. Its actual parsing is done by git-aggregator itself by using the [`autoaggregate` script](https://github.com/Tecnativa/docker-odoo-base/blob/master/bin/autoaggregate). Possibly you'll need to add an argument to that string that lets you specify the file you are aggregating (from custom or from auto).

build.d/300-oca-dependencies
PCatinean: Ah so I need to use the result from that method to find out the actual addons instead of just using the file itself. Any idea how I can debug easier maybe even use PDB to save time and not build everything from scratch to test the python script?

Yajo: Try running a project with scaffolding and everything, and mounting this script as a volume under it. Then run it manually inside the container.

build.d/300-oca-dependencies
PCatinean: I think I checked the output of this and it made the proper output i.e:

`target: origin $ODOO_VERSION `

Isn't that right?

Yajo: Yes, that's the proper output, although I'd be surprised if it gets there like that. I'd expect this line to be `'target': 'origin $ODOO_VERSION',` instead.

PCatinean:
I will resume work on this when I get a bit more time. I couldn't follow every reply on the mailing list regarding the pip installing method vs Doodba. This is still needed right?

Yajo:
Yes, as explained in https://github.com/OCA/runbot-addons/issues/144#issuecomment-338960929, I still see many benefits on addons.yaml system, I don't think it will fade away soon (although pip installing support is still there, of course)

Yajo:
According to https://github.com/OCA/runbot-addons/issues/144#issuecomment-339943623 (and mostly to the whole conversation) it seems OCA is not interested in this project, so please @PCatinean could you split this PR in 2?

1. Autofill `auto/repos.yaml` when a repo is declared in `custom/src/addons.yaml` but not in `custom/src/repos.yaml` by using `$DEFAULT_REPO_PATTERN`.
2. Implement `oca_dependencies.txt` recursive translation to this image's systems.

... Or otherwise just do step 1. Step 2 was a step needed to use Doodba in OCA. Since that's not going to happen, most likely it means that it goes to the collection of stuff that we just won't need. 🤔

PCatinean:
@Yajo I see, so basically in step 1 whenever an addon is declared in addons.yaml but it's not present in repos.yaml assume it's an oca repo and update repos.yaml accordingly?

Yajo:
yep!

PR #89, opened 2017-10-09
Change addons-install for a better CLI experience

Yajo:
It turns out for linters you possibly need full path of addon.

Current CLI was not very sane, changing it since it's a young feature that I don't expect to be used widely out there.


Yajo:
It seems the build is getting hit by https://github.com/travis-ci/travis-ci/issues/7401

Yajo:
🎉 This is a good step to get a doodba-quality-tools project up! (oops, spoiler!)

PR #93, opened 2017-10-22
Registering only valid Odoo modules from custom dir

PCatinean:
@Yajo fix for  https://github.com/Tecnativa/docker-odoo-base/issues/91

### DIFF
--- a/lib/odoobaselib/__init__.py
+++ b/lib/odoobaselib/__init__.py
@@ -42,6 +42,7 @@ def addons_config():
     """Yield addon name and path from ``ADDONS_YAML``."""
     config = dict()
     special_missing = {PRIVATE, CORE}
+    manifest_files = set(['__manifest__.py', '__openerp__.py'])
     try:
         with open(ADDONS_YAML) as addons_file:
             for doc in yaml.load_all(addons_file):
@@ -57,6 +58,15 @@ def addons_config():
                     for glob in addons:
                         logging.debug("Expanding glob %s", glob)
                         for addon in iglob(os.path.join(SRC_DIR, repo, glob)):
+                            if not os.path.isdir(addon):
+                                continue
+                            addon_files = set(os.listdir(addon))
+                            has_manifest = manifest_files & addon_files
+                            py_module = '__init__.py' in addon_files
+                            if not py_module or not has_manifest:
+                                logging.debug("Skipping '%s' as it is not a "
+                                    "valid Odoo module" % addon)
+                                continue
                             logging.debug("Registering addon %s", addon)
                             addon = os.path.basename(addon)
                             config.setdefault(addon, set())
--- a/lib/odoobaselib/__init__.py
+++ b/lib/odoobaselib/__init__.py
@@ -42,7 +42,7 @@ def addons_config():
     """Yield addon name and path from ``ADDONS_YAML``."""
     config = dict()
     special_missing = {PRIVATE, CORE}
-    manifest_files = set(['__manifest__.py', '__openerp__.py'])
+    manifest_files = ['__manifest__.py', '__openerp__.py']
     try:
         with open(ADDONS_YAML) as addons_file:
             for doc in yaml.load_all(addons_file):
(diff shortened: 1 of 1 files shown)


PCatinean:
Not sure what's going on with the last failing test

lib/odoobaselib/__init__.py
```diff
@@ -57,6 +58,15 @@ def addons_config():
...
                         logging.debug("Expanding glob %s", glob)
                         for addon in iglob(os.path.join(SRC_DIR, repo, glob)):
+                            if not os.path.isdir(addon):
+                                continue
+                            addon_files = set(os.listdir(addon))
+                            has_manifest = manifest_files & addon_files
+                            py_module = '__init__.py' in addon_files
+                            if not py_module or not has_manifest:
+                                logging.debug("Skipping '%s' as it is not a "
+                                    "valid Odoo module" % addon)
```
Yajo: Please do proper indentation here 😊

lib/odoobaselib/__init__.py
```diff
@@ -57,6 +58,15 @@ def addons_config():
                     for glob in addons:
                         logging.debug("Expanding glob %s", glob)
                         for addon in iglob(os.path.join(SRC_DIR, repo, glob)):
+                            if not os.path.isdir(addon):
+                                continue
+                            addon_files = set(os.listdir(addon))
+                            has_manifest = manifest_files & addon_files
+                            py_module = '__init__.py' in addon_files
```
Yajo: I think the `py_module` check can be misleading. What if a `setup` dir contains one of `__init__.py` files? `has_manifest` should be enough IMHO.

lib/odoobaselib/__init__.py
```diff
@@ -57,6 +58,15 @@ def addons_config():
                     for glob in addons:
                         logging.debug("Expanding glob %s", glob)
                         for addon in iglob(os.path.join(SRC_DIR, repo, glob)):
+                            if not os.path.isdir(addon):
+                                continue
+                            addon_files = set(os.listdir(addon))
```
Yajo: I feel that `if not (os.path.isfile(os.path.join(addon, "__manifest__.py")) or os.path.isfile(os.path.join(addon, "__openerp__.py"))):` would be faster, don't you think? 🤔

PCatinean: Indeed, right you are.

lib/odoobaselib/__init__.py
PCatinean: Indeed, I just thought that both conditions should be satisfied just like a regular Odoo module, though manifest should be ok. I did however check for both forms openerp and manifest so it keeps compatibility with V8.

lib/odoobaselib/__init__.py
```diff
@@ -58,14 +58,13 @@ def addons_config():
...
-                            addon_files = set(os.listdir(addon))
-                            has_manifest = manifest_files & addon_files
-                            py_module = '__init__.py' in addon_files
-                            if not py_module or not has_manifest:
-                                logging.debug("Skipping '%s' as it is not a "
-                                    "valid Odoo module" % addon)
+                            manifests = [
+                                os.path.join(addon, m) for m in manifest_files
+                            ]
+                            if not any([os.path.isfile(m) for m in manifests]):
```
Yajo: Remove the brackets please.

lib/odoobaselib/__init__.py
```diff
@@ -58,14 +58,13 @@ def addons_config():
...
+                                logging.debug(
+                                    "Skipping '%s' as it is not a valid Odoo "
+                                    "module" % addon)
```
Yajo: Replace `%` by `,` (a.k.a. `addon` should be function's 2nd argument).

lib/odoobaselib/__init__.py
```diff
@@ -58,14 +58,13 @@ def addons_config():
...
-                                continue
-                            addon_files = set(os.listdir(addon))
-                            has_manifest = manifest_files & addon_files
-                            py_module = '__init__.py' in addon_files
-                            if not py_module or not has_manifest:
-                                logging.debug("Skipping '%s' as it is not a "
-                                    "valid Odoo module" % addon)
+                            manifests = [
+                                os.path.join(addon, m) for m in manifest_files
+                            ]
```
Yajo: Replace `[...]` for `(...)`, to save some iterations.

lib/odoobaselib/__init__.py
```diff
@@ -57,6 +58,14 @@ def addons_config():
                     for glob in addons:
                         logging.debug("Expanding glob %s", glob)
                         for addon in iglob(os.path.join(SRC_DIR, repo, glob)):
+                            manifests = [
+                                os.path.join(addon, m) for m in manifest_files
+                            ]
+                            if not any([os.path.isfile(m) for m in manifests]):
```
Yajo: Remove brackets

lib/odoobaselib/__init__.py
```diff
@@ -57,6 +58,14 @@ def addons_config():
                     for glob in addons:
                         logging.debug("Expanding glob %s", glob)
                         for addon in iglob(os.path.join(SRC_DIR, repo, glob)):
+                            manifests = [
+                                os.path.join(addon, m) for m in manifest_files
+                            ]
```
Yajo: Replace brackets for parenthesis

PCatinean:
@Yajo not a problem, I get to learn more in the process and it should be just how you want it since it's your work after all. So never be afraid to keep insisting on my PR's I won't ever mind :smiley:

PR #94, opened 2017-10-25
Fix execution order of build.d scripts

Yajo:
They are now executed sorted by their alphabetical order, including files from both folders (common & custom).

Tests added.


Yajo:
The fix is OK, the failure is broken in master and is another bug.

PR #100, opened 2017-11-20
[IMP] Ensure dependencies installation replicability.

crimoniv:
Use `OrderedDict` instead of `dict` to ensure the same installation priority between builds (apt > gem > npm > pip) in [build.d/200-dependencies](/Tecnativa/docker-odoo-base/blob/master/build.d/200-dependencies).

This change also ensures that `pip install` dependencies' can be installed through Debian's `apt` (e.g.: `lxml` -> `libxml2-dev libxslt1-dev`).

### DIFF
--- a/lib/odoobaselib/installer.py
+++ b/lib/odoobaselib/installer.py
@@ -119,4 +119,3 @@ INSTALLERS = OrderedDict([
 def install(installer, file_path):
     """Perform a given type of installation from a given file."""
     return INSTALLERS[installer](file_path).install()
-
--- a/lib/odoobaselib/installer.py
+++ b/lib/odoobaselib/installer.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from collections import OrderedDict
 from os.path import exists
 from subprocess import check_call
 from odoobaselib import logging
@@ -107,14 +108,15 @@ class PipInstaller(Installer):
         return [self.file_path] if exists(self.file_path) else []
 
 
-INSTALLERS = {
-    "apt": AptInstaller,
-    "gem": GemInstaller,
-    "npm": NpmInstaller,
-    "pip": PipInstaller,
-}
+INSTALLERS = OrderedDict([
+    ("apt", AptInstaller),
+    ("gem", GemInstaller),
+    ("npm", NpmInstaller),
+    ("pip", PipInstaller),
+])
 
 
 def install(installer, file_path):
     """Perform a given type of installation from a given file."""
     return INSTALLERS[installer](file_path).install()
+
--- a/build.d/200-dependencies
+++ b/build.d/200-dependencies
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
 # Copyright 2017 LasLabs Inc.
 
-from os.path import join
+from os.path import basename, join, splitext
 from glob import glob
+import logging
 
 from odoobaselib import (
     AUTO_REQUIREMENTS,
@@ -25,3 +26,14 @@ for name in INSTALLERS:
     req_files.append(join(CUSTOM_DIR, 'dependencies', '%s.txt' % name))
     for req_file in req_files:
         install(name, req_file)
+
+# Sorted dependencies installation
+dep_files = sorted(glob(join(CUSTOM_DIR, 'dependencies', '[0-9]*-*')))
+for dep_file in dep_files:
+    root, ext = splitext(basename(dep_file))
+    # Get the installer (xxx-installer[-description][.ext])
+    installer = root.split('-', 2)[1]
+    if installer not in INSTALLERS:
+        logging.error("Unknown installer: %s", installer)
+        raise Exception
+    install(installer, dep_file)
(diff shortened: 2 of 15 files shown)


Yajo:
As you can see, apt build dependencies are installed before all of the others: https://github.com/Tecnativa/docker-odoo-base/blob/6cc9cb892ca03c5ae95c790171af5b05d3117bec/build.d/200-dependencies#L16-L19

So is this patch really fixing something, or is it just that you were trying to use `apt.txt` file for what you should have used `apt_build.txt`? 🤔

crimoniv:
The latter more than the former, I wrongly assumed that `apt_build.txt` was some kind of wrapper to allow building packages with` apt-build`.

Anyway, I think that having one `apt` packages' list –that, also, ensures being executed at the beginning–, feels more comfortable to me. What do you think?

Yajo:
The purpose of having split apt dependencies is that the build dependencies get uninstalled after installing all other dependencies, resulting in a thinner image, so that's not gonna change.

About having deterministic builds... It seems a good thing, but it could be wrong too, because I think that the order of installation of all other package systems should not matter. Otherwise one could ask: why do we install always npm packages before pip ones? And I see no reason on why to choose a specific order. @lasley what do you think about this?

crimoniv:
> The purpose of having split apt dependencies is that the build dependencies get uninstalled after installing all other dependencies, resulting in a thinner image, so that's not gonna change.

I see what you mean, I didn't known about this snippet –that explains the purpose of `apt_build.txt` and why it shouldn't be removed–:
https://github.com/Tecnativa/docker-odoo-base/blob/6cc9cb892ca03c5ae95c790171af5b05d3117bec/build.d/900-dependencies-cleanup#L11-L14

In any case, I still believe that it is desirable for a building process to be replicable -with or without apparent improvements-. Maybe the order could be changed to be more _python-oriented_ (`apt` > `pip` > `npm` > `gem`).

As another possible use case of this change, `pip` dependencies could be satisfied without the need to compile them through `pip`, saving some installation time. E.g., if `python-numpy` was installed through `apt`, `matplotlib` will have this requirement satisfied and, therefore, it won't be compiled.

lasley:
> Otherwise one could ask: why do we install always npm packages before pip ones? And I see no reason on why to choose a specific order. @lasley what do you think about this?

Hmmmm but on the flipside, what if an apt is required to both build and run a package dependency? Pretty rare, but I assume there could be such a thing - basically a Python/Node package that has to interact with a CLI due to lack of development headers.

But then we run into the question of what exactly needs to be installed first? Apt is a given as first, but there's an equal chance of a Python something using a Gem, or a Gem using something Node. Or maybe this isn't an issue, because the package will be there at the end anyways and it wasn't actually a dependency?

> As another possible use case of this change, pip dependencies could be satisfied without the need to compile them through pip, saving some installation time. E.g., if python-numpy was installed through apt, matplotlib will have this requirement satisfied and, therefore, it won't be compiled.

I take reservation to this (although it's obviously whatever you want to do for dependency management). I try to run on the absolute most updated software I can, but the pre-built dependencies are fairly out of date.

Yajo:
Then one good solution would be to have a build argument for this.

Add `ARG DEPS_ORDER=apt,npm,gem,pip` to the `Dockefile`s. Let's default to having apt first and pip last, since those 2 seem to be the most interesting for us.

Of course, apt_build still must stay 1st of all and autoremove itself.

This way we leave sane deafults while actually providing some value on allowing the subproject pinning the build order.

The hardest part is how to test, though.

crimoniv:
How about changing to a naming convention that specifies the installer to use in addition to the order? E.g.:
```
odoo/custom/dependencies/000-apt.txt
odoo/custom/dependencies/005-apt-with-description.txt
odoo/custom/dependencies/010-apt-python-reqs.txt
odoo/custom/dependencies/100-pip-more-python-reqs.txt
odoo/custom/dependencies/200-npm-my-node-packages.txt
...
```

lasley:
That's a damned good idea @crimoniv

Yajo:
It seems legit, please proceed, and remember to keep backwards compatibility. 😉

crimoniv:
Apologies for the delay!

I've updated the PR with the latest changes, always ensuring backwards compatibility and including related tests. Let me know what do you think!

lib/odoobaselib/installer.py
```diff
@@ -107,14 +108,15 @@ def requirements(self):
...
+    ("gem", GemInstaller),
+    ("npm", NpmInstaller),
+    ("pip", PipInstaller),
+])
 
 
 def install(installer, file_path):
     """Perform a given type of installation from a given file."""
     return INSTALLERS[installer](file_path).install()
+
```
Yajo: 🔥

Yajo:
The error was #109, which is now fixed. Rebase please.

crimoniv:
Done! We've successfully run all tests locally, so hopefully it works now :crossed_fingers: 

P.S. We've also replaced an `f-string` with an `str.format` to add Python 3.5 compatibility (the current Debian stable Python version), hope you don't mind.

PR #105, opened 2017-12-26
[IMP] Downgrade to wkhtmltopdf 0.12.1 in Odoo 8.0

crimoniv:
The included wkhtmltopdf version (0.12.4) has some rendering differences in relation to the version 0.12.1. Prefer the latter in Odoo 8.0 as indicated in the [Odoo Wiki](https://github.com/odoo/odoo/wiki/Wkhtmltopdf).

### DIFF
--- a/8.0.Dockerfile
+++ b/8.0.Dockerfile
@@ -51,8 +51,9 @@ ONBUILD RUN /opt/odoo/common/build && sync
 ONBUILD USER odoo
 
 ARG PYTHONOPTIMIZE=2
-ARG WKHTMLTOPDF_VERSION=0.12.4
-ARG WKHTMLTOPDF_CHECKSUM='049b2cdec9a8254f0ef8ac273afaf54f7e25459a273e27189591edc7d7cf29db'
+ARG WKHTMLTOPDF_VERSION=0.12.1
+ARG WKHTMLTOPDF_CHECKSUM='3ae78d7f7323f6bbc76238eb548729f17ab0db5bed59455a023b9c93fb0c35c4'
+ARG WKHTMLTOPDF_URL=http://nightly.odoo.com/extra/wkhtmltox-0.12.1.2_linux-jessie-amd64.deb
 ENV DEPTH_DEFAULT=1 \
     DEPTH_MERGE=100 \
     EMAIL=https://hub.docker.com/r/tecnativa/odoo \
@@ -108,10 +109,12 @@ RUN ln -s /usr/bin/nodejs /usr/local/bin/node \
     && rm -Rf ~/.npm /tmp/*
 
 # Special case for wkhtmltox
-RUN curl -SLo wkhtmltox.tar.xz https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/${WKHTMLTOPDF_VERSION}/wkhtmltox-${WKHTMLTOPDF_VERSION}_linux-generic-amd64.tar.xz \
-    && echo "${WKHTMLTOPDF_CHECKSUM}  wkhtmltox.tar.xz" | sha256sum -c - \
-    && tar --strip-components 1 -C /usr/local/ -xf wkhtmltox.tar.xz \
-    && rm wkhtmltox.tar.xz \
+RUN curl -SLo wkhtmltox.deb "${WKHTMLTOPDF_URL}" \
+    && echo "${WKHTMLTOPDF_CHECKSUM}  wkhtmltox.deb" | sha256sum -c - \
+    && dpkg --force-depends -i wkhtmltox.deb \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends -f \
+    && rm -Rf /var/lib/apt/lists/* wkhtmltox.deb \
     && wkhtmltopdf --version
 
 # Other facilities


pedrobaeza:
:-1: We have used this version since a long time without issues. Please indicate concrete problems, but don't follow that guide as the principle.

lasley:
Agreed on 👎  for identical reasoning as Pedro. 

And as a general note, I do not recommend treating the Odoo docs as gospel. I find that in practical usage, many of them are wrong or recommending bad practices (such as the use of software from >3 years ago).

crimoniv:
The same report -`sale.report_saleorder`- in `0.12.4` (left) and `0.12.1` (right). The differences are very subtle but enough to broke some of our custom formats adapted to specific printers.
sale report_saleorder

pedrobaeza:
Yeah, but that's a problem if you come from the other version, not if the reports are adapted to this one. I recommend you to adapt the reports to the new version. Sooner or later you will need to do it.

crimoniv:
Well.. The recommended version has always always been the 0.12.1 until the 0.12.4 was released on November 22, 2016, and Odoo 8.0 was released on September 18, 2014.

Please note that this is only applicable to Odoo 8.0, which has already reached its EOL.

pedrobaeza:
I repeat that we have been working with that new version since then without any other issue except the little font discrepancy, and we have adapted reports that way. With that, when migrating to a new version of Odoo, you don't have to make that step.

lasley:
In the instance where a customer is still using v8, I would assume they're cheaping out on engineering labor & probably wouldn't want to pay for the upgrade. 

> Please note that this is only applicable to Odoo 8.0, which has already reached its EOL.

This change applies to version 9 and 10 as well. Those Dockerfiles are just symlinks. I think the effort required to maintain another Dockerfile is quite large and should be avoided (this is already being done for v11).

Why not just switch the `WKHTMLTOPDF_VERSION` and `WKHTMLTOPDF_CHECKSUM` in your build to match the version you need?

crimoniv:
> This change applies to version 9 and 10 as well. Those Dockerfiles are just symlinks.
My bad! Sorry, I was not aware of that.

> I think the effort required to maintain another Dockerfile is quite large and should be avoided (this is already being done for v11).
We wouldn’t mind to create and maintain the 8.0 Dockerfile, if you agree.

> Why not just switch the WKHTMLTOPDF_VERSION and WKHTMLTOPDF_CHECKSUM in your build to match the version you need?
The installer provided by Odoo was specifically compiled for Debian 8.0, and there is no other compatible binary.

crimoniv:
We have created a new `8.0.Dockerfile` in the [same branch](https://github.com/diagramsoftware/docker-odoo-base/tree/odoo8-downgrade-wkhtmltox-version), just in case you reconsider reopening this PR.

Thank you anyway for your attention @lasley!

lasley:
@crimoniv - I'm just a contributor, you'd need to petition to Tecnativa to have it be reopened. 

My standpoint is that we should be able to support the recommended version (0.12.1.x), but we shouldn't default to it. This would allow us to keep the same Dockerfiles, which otherwise become a real maintenance burden as time goes forward. I'm just a contributor though, and my say does not hold as much weight here as @Yajo or @pedrobaeza. 

Note - while you're offering to contribute the original Dockerfile, I can almost guarantee you are not going to put your neck out for the ongoing maintenance of this Dockerfile. For the most part, any change in one file needs to be made in all three and subsequently tested. I'm already personally having troubles remembering to make changes in the two files we already have. I don't think that adding another is the way to go.

pedrobaeza:
We don't want another Dockerfile indeed. You can trick your scaffolding for installing another wkhtmltopdf version any way if you don't want the latest one.

lasley:
IMO it would be nice for a supported way to run the outdated version though. It's technically the recommended version & is already on pre-deployed instances. 

Forcing the new version makes it harder for a smooth adoption of this base, because unknown amounts of dev effort needs to be put in to the XML migrations & customers may not want to pay for something that was already working. 

This could make or break the decision for an Odoo implentor to adopt this deployment method organization-wide. From my end, hosting maintenance effort is rolled into my Odoo hosting packages. Assuming this is the case for others', it means that development effort not previously factored into the cost would need to go into pre-existing instances in order to migrate to this new deploy method.

It's definitely possible to have this work via a scaffold, and in fact, it's very easy. I just wonder how many people would be forced to do this in their own scaffolds? I haven't figured out a way to chain scaffolds yet, so this means that the same code would need to be recreated and maintained individually per customer scaffold. 

Again, not a problem for me personally, just elaborating on possible reasoning to why this should be included in base if we could find a happy medium.

pedrobaeza:
Well, if someone makes the work...

crimoniv:
@lasley

> Note - while you're offering to contribute the original Dockerfile, I can almost guarantee you are not going to put your neck out for the ongoing maintenance of this Dockerfile.

We are really concerned about keeping a functional Odoo 8.0 but don't worry, I completely understand your stance on this subject.

@pedrobaeza

> We don't want another Dockerfile indeed. You can trick your scaffolding for installing another wkhtmltopdf version any way if you don't want the latest one.

We have already tried this approach, but the non-cacheable increase of 60~80MB in the resulting image –e.g., whenever the addons.yaml/repos.yaml pair is modified–, isn't worth it.

---

We have finally reached this approach, which we hope that fulfills this _happy medium_: https://github.com/diagramsoftware/docker-odoo-base/commit/5ac820ccf74bf3ad34fe775faf957b602f2b235a

pedrobaeza:
When @Yajo comes back from holidays will say if technically is correct.

Yajo:
I have to admit I understand your concerns, @crimoniv, but there are reasons why this PR cannot be accepted:

- It would break all preexisting v8 deployments in the inverse way it's breaking yours: People who already upgraded their reports to make them work with newer wkhtmltopdf versions (which, sooner or later, you'll have to do) would see them broken again, and would need to re-downgrade them. That's just not acceptable, I'm sure you understand that.
- If you clone the repository and build your own image, it's just a matter of adding a couple of build arguments and customizing your base build: https://github.com/Tecnativa/docker-odoo-base/blob/318cf1a65c6977418299e7936d68fcde72c19ef3/8.0.Dockerfile#L54-L55
- If you decide to stick with our official base image, [you can just drop any executable into your scaffolding's `build.d` directory](https://github.com/Tecnativa/docker-odoo-base#optodoocustombuildd) to modify that project's build.
- There are stability problems with those older whktmltopdf versions that got fixed in newer. Can't remember a specific case right now, but if we upgraded it, it was for that reason mainly.
- The extra maintenance burden also counts.

So, summarizing, your situation is already covered because the tool already provides the hooks you need to customize it, but it's not going to be the default.

Sorry friend, but thanks! ☺️

lasley:
> If you clone the repository and build your own image, it's just a matter of adding a couple of build arguments and customizing your base build:

Actually @Yajo I have to disagree here. Take a look at the files in the GitHub releases for 12.1.x and compare to the newer ones - there are not precompiled binaries. You need an entirely separate URI for these, and the owner of the repo has explicitly declined to put the compiled versions of the 12.1.x on GitHub releases. 

But I still agree with @Yajo on the overall. My question is if this would be accepted if it were a build script that was triggered by an Environment variable? Default to the 12.4.x, but allow for a walk backwards if necessary. It really just sucks that the maintainer refuses to align the release file type - the switch up is drastic.

Yajo:
Ah, good point!

Well, as long as the PR is opened with proper tests and no official 0.12.1 builds are expected, that would be accepted.

However if the difference is so big, I'd still encourage the addition of a build.d script, at the cost of those extra MB. That script could be contributed to the wiki instead.

PR #107, opened 2018-01-03
[NEW] convert smtpfake to mailhog catcher

### LINKED ISSUE #101: Use a better fakesmtp
Current fakesmtp just logs "a mail has been sent". Not very useful for development.

https://github.com/mailhog/MailHog seems to be a better default for that purpose. Maybe there are others? They have an official docker image; it should be easy to plug in the scaffolding.

njeudy:
Use mailhog as smtp catcher for dev environment

- fix #101 

WIP until:

- [x] #110

**howto**: when docker-compose -f devel.yaml up, go to http://localhost:8025 to have mailhog.

### DIFF
--- a/devel.yaml
+++ b/devel.yaml
@@ -54,7 +54,6 @@ services:
             file: common.yaml
             service: smtpfake
         ports:
-            - "127.0.0.1:1025:1025"
             - "127.0.0.1:8025:8025"
 
     wdb:
--- a/devel.yaml
+++ b/devel.yaml
@@ -51,6 +51,8 @@ services:
         extends:
             file: common.yaml
             service: smtpfake
+        ports:
+            - "127.0.0.1:8025:8025"
 
     wdb:
         image: yajo/wdb-server
--- a/devel.yaml
+++ b/devel.yaml
@@ -17,7 +17,6 @@ services:
             ADMIN_PASSWORD: admin
             PGDATABASE: devel
             PYTHONOPTIMIZE: ""
-            SMTP_SERVER: "smtp"
             SMTP_PORT: "1025"
             # You want demo data for development
             WITHOUT_DEMO: "false"
--- a/devel.yaml
+++ b/devel.yaml
@@ -52,6 +52,7 @@ services:
             file: common.yaml
             service: smtpfake
         ports:
+            - "127.0.0.1:1025:1025"
             - "127.0.0.1:8025:8025"
 
     wdb:
--- a/devel.yaml
+++ b/devel.yaml
@@ -17,6 +17,8 @@ services:
             ADMIN_PASSWORD: admin
             PGDATABASE: devel
             PYTHONOPTIMIZE: ""
+            SMTP_SERVER: "smtp"
+            SMTP_PORT: "1025"
             # You want demo data for development
             WITHOUT_DEMO: "false"
         ports:
--- a/common.yaml
+++ b/common.yaml
@@ -30,7 +30,7 @@ services:
             - db:/var/lib/postgresql/data:z
 
     smtpfake:
-        image: tecnativa/smtp-sink
+        image: mailhog/mailhog
 
     smtpreal:
         image: tecnativa/postfix-relay
--- a/test.yaml
+++ b/test.yaml
@@ -8,6 +8,8 @@ services:
             PGDATABASE: test
             # You may want demo data for testing
             WITHOUT_DEMO: "false"
+            SMTP_SERVER: "smtp"
+            SMTP_PORT: "1025"
         depends_on:
             - db
             - smtp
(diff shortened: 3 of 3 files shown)


njeudy:
@Yajo my first contrib for this new year :) 

And have some question why didn't you use inverse_proxy for dev env ? using 127.0.0.1 is limited to one project at time .. I use inverse_proxy in dev env and it is more accurate I think ..

Yajo:
🎉 Great small contribution pal!
About your question of using localhost in devel, it was with the idea of having a local environment set up more easily. If we use inverse_proxy, we do not only need to have a local traefik running, we also need to patch `/etc/hosts` for each new project.

It seemed easier. However further discussion can have place in a separate issue if you want, but let's not drift away from this one's purpose, ok?

I'll test this locally ASAP, Thanks!

Yajo:
The only problem I currently see is that the fake SMTP port (1025) differs from the one open at the real SMTP (25), but I had some problems when opening it in Mailhog. Namely, it logged this:

    2018/01/05 12:36:20 [SMTP] Error listening on socket: listen tcp 0.0.0.0:25: bind: permission denied

Not sure right now about the problem, but I guess it's not much critical as long as we know it.

Please @njeudy approve my changes too to merge.

pedrobaeza:
But that means to change each time the port on the mail server configuration? I would prefer not to do it...

Yajo:
The problem would be when restoring a production db in a test/devel environment.

Apart from that, you always need to change the mail server configuration nowadays, because odoo's defaults don't work at all.

Not sure if that's what you meant...

pedrobaeza:
localhost and 25 (defaults) doesn't work?

Yajo:
No, you have to change it to `smtp:25`

pedrobaeza:
Uhm, I see...

Yajo:
So? Then not much more problem to change to `smtp:1025`, isn't it?

Well, I can investigate more why port 25 wasn't able to bind, but I guess it's a very standard port and can hit problems on more than one environment...

pedrobaeza:
Yes, it's reasonable, but please add the documentation

njeudy:
@Yajo @pedrobaeza I check this

You can not bind port under 1024 with docker but I think you don't need to bind port in host because you only need http access ..

njeudy:
Web can also add script in entrypoint.d to remove internal odoo config ?

psql -c "delete from ir_mail_server;" -d "${PGDATABASE}"
psql -c "delete from fetchmail_server;" -d "${PGDATABASE}"

njeudy:
@pedrobaeza what kind of doc you need ?

https://github.com/mailhog/MailHog ?

pedrobaeza:
I mean in the docker project for knowing how to configure the mail server.

njeudy:
@pedrobaeza @Yajo so sorry, I use a bit different version of docker-odoo-base and did'nt notice my merge request was not complete.

see #110 before and this merge request after.

@pedrobaeza : can you explain me what documentation you need ? technical setup for mailhog + odoo ? or functional setup in odoo ?

devel.yaml
```diff
@@ -51,6 +53,9 @@ services:
         extends:
             file: common.yaml
             service: smtpfake
+        ports:
+            - "127.0.0.1:1025:1025"
```
njeudy: @Yajo think you don't need this. only http interface to 8025 is needed. 1025 port is internal.

pedrobaeza:
Maybe document it here... https://github.com/Tecnativa/docker-odoo-base/blob/master/README.md#development

@Yajo can say which is the best part

njeudy:
@pedrobaeza ok, will update my commit on docker-odoo-base merge request with some info to setup mailhog / odoo 

@Yajo just say to me where is it the best for you ?

devel.yaml
Yajo: Yeah, in my mind it seemed useful, so we'd be able to connect i.e. Outlook (which you know how badly renders emails) to the local port and see the email, but actually it's just an SMTP port (not IMAP/POP3) so it's useless as you say.

Yajo:
Let's continue each conversation in its PR 😉

Yajo:
Now I know why it didn't have permission to listen on `0.0.0.0:25`. It's a privileged port, and [MailHog runs as the mailhog user by default](https://github.com/mailhog/MailHog/blob/1bd561aa34bf68e52eddcdab46c45bdaa576338c/Dockerfile#L27).

By just adding `user: root` to the smtpfake definition, it works and shares the same port.

However, I added tests to #110 and it seems they should work, so I think I still prefer that approach. If we can have the same functionality giving less privileges, it's a good thing IMHO.

Yajo:
This one is ready to merge. @njeudy do you mind if I squash all of this into one commit?

njeudy:
@Yajo non problem for me :)

PR #108, opened 2018-01-04
Make local binaries always findable

Yajo:
It turns out that without this patch, `subprocess.call(["flake8", "--version"])` from Python wouldn't work if it were installed just for the user.


11.0.Dockerfile
```diff
@@ -57,7 +57,7 @@ ENV DEPTH_DEFAULT=1 \
     LC_ALL=C.UTF-8 \
     NODE_PATH=/usr/local/lib/node_modules:/usr/lib/node_modules \
     OPENERP_SERVER=/opt/odoo/auto/odoo.conf \
-    PATH="~/.local/bin:$PATH" \
+    PATH="/home/odoo/.local/bin:$PATH" \
```
pedrobaeza: Hardcoding a path is not a good idea IMO.

Yajo: It's not a problem when all project is hardcoded. That's the beauty of making decisions. 😊

pedrobaeza: But why it doesn't work? If you make `run` with the proper user (the same that installs locally the bin file)...

Yajo: Because the `~` is hardcoded into the `$PATH` variable, instead of being expanded. This makes some environments to detect it, such as bash, but some others to don't detect, such as python.

You can do the simple test in any project:

1. `docker-compose run --rm odoo bash -c 'pip install -q --user flake8 && which flake8'`
2. `docker-compose run --rm odoo bash -c 'pip install -q --user flake8 && PATH="/home/odoo/.local/bin:$PATH" which flake8'`

You'll see only the 2nd works

Yajo:
Well, the CI ❌ is #109, not related, and this is blocking doodba-qa, so I merge.

11.0.Dockerfile
pedrobaeza: I think there's a method in `os.path` for unfolding that symbols.

Yajo: Possibly, but that's not a problem of MQT, which is where we'd have to fix it if we wanted to avoid this.

PR #114, opened 2018-01-10
[ADD] Stall startup until another host dies

lasley:
* Add option to stall startup until a defined host stops responding to pings.

This was a feature from #41 that didn't make it in with the new implementation. 

It's something I find as a requirement though, because it allows for you to orchestrate back-to-back operations by chaining containers that exit when completed. My entire CD pipeline is based on this, interestingly enough.


entrypoint.d/90-wait-no-host
```diff
@@ -0,0 +1,16 @@
...
+log INFO "Waiting for the ${WAIT_NO_HOST} container to stop."
+
+while ping -c1 "${WAIT_NO_HOST}" &> /dev/null; do
+    log DEBUG "${WAIT_NO_HOST} responded to ping."
+    sleep 1
+done
+
+log INFO "${WAIT_NO_HOST} has stopped responding; resuming startup."
+
+exit 0
```
Yajo: This is implicit, you don't need it.

entrypoint.d/90-wait-no-host
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+if [ -z "${WAIT_NO_HOST}" ]; then
+    exit 0
+fi
+
+log INFO "Waiting for the ${WAIT_NO_HOST} container to stop."
```
Yajo: This could be or not a container. Better say "host"

lasley:
Thanks for the review @Yajo - comments attended to.

Regarding the test - I agree we need one. I'm having a bit of trouble on it conceptually though. So my obvious thought is:

* Start some dummy container in the docker-compose
* Dummy container waits for postgres host to become available (that entrypoint is first, so the host would be waiting otherwise)
* Dummy container waits an indeterminate time for Odoo to boot and install the first db, but Odoo shouldn't actually be doing this
* Dummy container curls Odoo container after indeterminate time to verify the container is not actually continued
* Dummy container exits with a status code indicating the test response

But the issues:
* The indeterminate amount of time is a long time to stall tests for no reason. First Odoo boot is a solid minute at least.
* I don't see a way to read the exit codes from the dummy container in the tests.

The above rules my theory out, primarily due to the first point. For a new theory I'm playing with, I have questions that I haven't been able to answer myself yet:
* Can I read the Odoo container logs from our tests?
* Can I control the containers from our tests? (to stop the dummy)

Yajo:
Before answering your questions, I just think you're getting it the hard way, let's simplify it:

- Add new test scaffolding with just 2 containers: `odoo` and `blocker`.
- In the `odoo` one, add env `WAIT_DB=false` and `WAIT_NOHOST=blocker`.
- Also there, add a lower-level entrypoint that creates a file if `blocker` is alive.
- In the `blocker` one, add command `sleep 10 && true` (10 sec seems sane enough).
- In the test itself, check that the first entrypoint created the expected file.

The result will be:

1. `blocker` is booted
2. `odoo` is booted
2. 1st entrypoint creates file
2. 2nd entrypoint waits 10s
2. command checks file was created. If the command runs, it means both entrypoints passed.

Do you like it? 😊

Yajo:
This is very outdated and stale. Closing.

PR #117, opened 2018-01-25
[DCK] Remove inbox containers, use internal networks

### LINKED ISSUE #98: Use --internal with sane defaults in scaffoldings
It's interesting that in the scaffolding there has been the `fakeinbox` and `fakesmtp` containers made to isolate restored backups in devel/test environments from making requests to the outside world, specifically when attempting to contact servers.

However, it turns out that since Docker 1.10, we can create networks with the `--internal` flag, which restricts access to the outside world perfectly fine (see https://github.com/moby/moby/pull/19276).

We can use that instead of using some of the proxies. It should be easier to understand for newcomers, and easier to maintain for oldcomers.

Yajo:
The inbox proxies were a trick to isolate POP3/IMAP (or any other TCP) connections when a database got restored into a non-production environment.

Since Docker 1.10, internal networks are natively supported, which means that any in/out communication outside the container stack gets automatically blocked.

From now on, you can just configure incoming mail as usual in your Odoo instances with the safety that those incoming mails will not be read outside the production environment.

If you still need outside world communication for devel or test environments, you can enable it by switching the internal boolean, or by adding a public network with public access and attaching it to the containers that need it (usually just Odoo).

Note that for this to have a real effect, the `inverseproxy_shared` network must be internal too in test environments.

Fixes #98

### DIFF
--- a/setup-devel.yaml
+++ b/setup-devel.yaml
@@ -20,6 +20,8 @@ services:
                 PIP_INSTALL_ODOO: "false"
                 CLEAN: "false"
                 COMPILE: "false"
+        networks:
+            - public
         volumes:
             - ./odoo/custom/src:/opt/odoo/custom/src:rw,z
         environment:
@@ -30,3 +32,6 @@ services:
             UMASK: "$UMASK"
         user: root
         entrypoint: autoaggregate
+
+networks:
+    public:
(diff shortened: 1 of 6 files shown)


setup-devel.yaml
```diff
@@ -20,6 +20,8 @@ services:
                 PIP_INSTALL_ODOO: "false"
                 CLEAN: "false"
                 COMPILE: "false"
+        networks:
+            - public
```
pedrobaeza: Public for this?

Yajo: Yes, in this case the running container needs to connect to the network to download git code from github usually.

PR #122, opened 2018-02-02
[NEW] add db_port config parameters

njeudy:
- Some time postgres is not on 5432 port (mostly cloud db storage)


njeudy:
@Yajo done with test :) !!

PR #126, opened 2018-02-21
Proxy for Odoo in devel

### LINKED ISSUE #120: Cannot access ports in devel.yaml
It seems that #117 raised yet another issue: cannot access ports in `devel.yaml`.

I opened https://github.com/moby/moby/issues/36174 to track it, docker-side.

For now, the fastest workaround would be to **set `internal: false` in `devel.yaml` and then `docker-compose down` and `docker-compose up -d` again**.

I'll investigate a better workaround when I have time.

Yajo:
To fix #120:

- Add a new "public" network, which has no restriction
- Bind wdb and smtp containers to that network, since they have no data leakage possibility
- Add a new odoo_proxy service: just a blind netcat that allows external connections to odoo
- Odoo and db are still isolated and cannot access the public network, so the isolation features remain untouched.

### DIFF
--- a/devel.yaml
+++ b/devel.yaml
@@ -2,7 +2,7 @@ version: "2.1"
 
 services:
     odoo_proxy:
-        image: alpine
+        image: tecnativa/whitelist
         depends_on:
             - odoo
         networks: &public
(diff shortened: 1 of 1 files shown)


Yajo:
FTR, while doing this research, I found a way to whitelist apis in test environments:

```yaml

    pypi_proxy:
        image: alpine
        depends_on:
            - odoo
        networks:
            default:
                aliases:
                    - pypi.python.org
            public:
        command:
            - sh
            - -c
            - apk add --no-cache bind-tools && nc -lkp 443 -e nc "$$(dig +short pypi.python.org @8.8.8.8 | sed 's/\.$$//' | head -n 1)" 443
```

This is obviusly not much easy to understand, I'm creating a new image in our stack for this, and I'll update this PR to use it.

PR #127, opened 2018-02-22
[FIX] Freeze psycopg2 library

crimoniv:
The current `psycopg2` version (`2.7.4` as of today) shows a **UserWarning** about future packaging changes. Force the version found in the `requirements.txt` from **OCA/OCB**.

This also ensures that the package `psycopg2` won't be reinstalled when enabling `AUTO_REQUIREMENTS` if the `requirements.txt` is synced with the **OCB**'s one.

The log in question is:
> /usr/local/lib/python2.7/dist-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
  """)

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -39,18 +39,12 @@ else
     pip_deps="psutil==4.3.1 pydot==1.2.3"
 fi
 
-# psycopg2 will be installed later
-sed -ir 's/psycopg2/#\0/' $reqs
-
 optimize="$PYTHONOPTIMIZE"
 if [ $PYTHONOPTIMIZE -gt 1 ]; then
     export PYTHONOPTIMIZE=1
 fi
 pip install --no-cache-dir $pip_deps
 export PYTHONOPTIMIZE="$optimize"
-
-# Security upgrades
-pip install --no-cache-dir "psycopg2>=2.7" # ODOO-SA-2017-06-15-1
 
 # Build and install Odoo dependencies with pip
 pip install --no-cache-dir --requirement $reqs


bin/install.sh
```diff
@@ -50,7 +50,11 @@ pip install --no-cache-dir $pip_deps
 export PYTHONOPTIMIZE="$optimize"
 
 # Security upgrades
-pip install --no-cache-dir "psycopg2>=2.7" # ODOO-SA-2017-06-15-1
+if [ "$ODOO_VERSION" == "8.0" ] || [ "$ODOO_VERSION" == "9.0" ]; then
+    pip install --no-cache-dir "psycopg2==2.7.1"
+else
+    pip install --no-cache-dir "psycopg2==2.7.3.1"
+fi
```
Yajo: Instead of doing this, please just remove the separate psycopg2 installation stuff (see line 43 too).

This was a patch referring to https://github.com/odoo/odoo/issues/17914 because they didn't update upstream `requirements.txt` files.

After awhile, I'm now seeing that https://github.com/OCA/OCB/commit/6f8efea28cd29fd973ac6a50d7b93f0d73a725c6 is found in 8-11 branches, so we don't need to patch anything downstream now.

crimoniv: You're right, it doesn't make sense to ignore the version from `requirements.txt` and manually install the exact same version. Changing it right now!

crimoniv:
Done :ok_hand:

PR #130, opened 2018-03-22
New update mechanism landed in module_auto_update

Yajo:
This should keep backwards compatibility wherever needed, while working with the newest features introduced in https://github.com/OCA/server-tools/pull/1190 wherever available. Deprecated features get autodisabled.

### DIFF
--- a/bin/autoupdate
+++ b/bin/autoupdate
@@ -1,5 +1,20 @@
 #!/usr/local/bin/python-odoo-shell
+import os
+
 
 # Note: ``module_auto_update`` must be installed in Odoo for this to work.
-env["base.module.upgrade"].upgrade_module()
+try:
+    env["ir.module.module"].upgrade_changed_checksum
+except AttributeError:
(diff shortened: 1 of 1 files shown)


bin/autoupdate
```diff
@@ -1,5 +1,20 @@
...
-env["base.module.upgrade"].upgrade_module()
+try:
+    env["ir.module.module"].upgrade_changed_checksum
+except AttributeError:
+    env["base.module.upgrade"].upgrade_module()
+else:
+    # Disable deprecated stuff
+    env["ir.config_param"].set_param(
+        "module_auto_update.enable_deprecated",
+        "0",
```
Yajo: I'm not sure if we should do this here or leave existing implantations to decide. That wouldn't harm us, I just wanted to publish this to hear others' opinions.

pedrobaeza: Well, I think it's not too problematic to leave this way.

PR #136, opened 2018-04-20
Support $INITIAL_LANG

Yajo:
Lets automated deployments of new databases with only one language that is not en_US.


pedrobaeza:
Have you put in our scaffolding the Spanish initial language?

Yajo:
Yes

PR #140, opened 2018-05-02
Upgrade to Node 6.x (LTS) and Less 2.x in all versions

Yajo:
This should remove some warnings from builds like these:

```
WARN engine less@3.0.2: wanted: {"node":">=4"} (current: {"node":"0.10.29","npm":"1.4.21"}) 
```

Also I hope it adds a workaround for one problem found, related to https://github.com/odoo/odoo/pull/24246, which rendered a useless website with:

    The "--no-js" argument is deprecated, as inline JavaScript is disabled by default. Use "--js" to enable inline JavaScript (not recommended).
    
    /usr/local/lib/node_modules/less/node_modules/request/node_modules/hawk/node_modules/boom/lib/index.js:5
    const Hoek = require('hoek');
    ^^^^^
    FileError: optional dependency 'request' required to import over http(s)
    in - on line 342, column 1:
    341 // General styles
    342 @import url(//fonts.googleapis.com/css?family=Open+Sans:400,400italic);
    343

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -96,7 +96,7 @@ RUN apt-get -qq update \
         bzip2 ca-certificates curl gettext-base git gnupg2 nano \
         openssh-client postgresql-client telnet xz-utils \
     && curl https://bootstrap.pypa.io/get-pip.py | python3 /dev/stdin \
-    && curl -sL https://deb.nodesource.com/setup_8.x | bash - \
+    && curl -sL https://deb.nodesource.com/setup_6.x | bash - \
     && apt-get install -yqq nodejs \
     && apt-get -yqq purge python2.7 \
     && apt-get -yqq autoremove \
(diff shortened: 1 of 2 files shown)


crimoniv:
We've also come across with this issue and finally opted to [froze the `less.js` version](https://github.com/diagramsoftware/docker-odoo-base/commit/c62261a7f55eb9b295627bddeba7fd8c04b1e1cb) to avoid breaking the [Odoo 8.0 themes compatibility](https://www.odoo.com/documentation/8.0/setup/install.html#installing-dependencies). Feel free to reuse the code if it seems good to you!

crimoniv:
> I tested and in effectively: that bug only happens with less >= 3.0.

I'm glad it also works in Odoo 9.0! I would have liked to help you but we've never worked with themes in Odoo >= 9.0

> Let's see what happens now... I can't imagine how could anybody install sanely Odoo without docker. 😕

Yeah... Odoo installation is a bit trickier sometimes. Thank God Docker simplifies this process a lot!

PR #142, opened 2018-05-18
[FIX] setting environment variable UNACCENT to false

ap-wtioit:
when setting environment variable UNACCENT to false this should be also reflected in odoo.conf

Previously we would get a warning in the logfile of odoo if we started the docker container with environment UNACCENT: "false":
init The option --unaccent was given but no unaccent() function was found in database.

This happened because 30-unaccent-install did skip installing to db but /opt/odoo/auto/odoo.conf was already generated in build with arg UNACCENT=true

Info @wt-io-it

### DIFF
--- a/entrypoint.d/35-unaccent-config
+++ b/entrypoint.d/35-unaccent-config
@@ -1,0 +1,4 @@
+#!/bin/bash
+set -e
+# correct unaccent in config to the given environment value
+sed -i 's/\(unaccent = \)\(.*\)$/\1 '$UNACCENT'/' /opt/odoo/auto/odoo.conf


Yajo:
Sorry, generating config at build time is deprecated since #38. It's now done in the entrypoint.

The problem you say is because to install unaccent you need a database to exist, but in the 1st boot, Odoo hasn't still created the DB at the time when the entrypoint is executed.

It will simply work if you reboot the container after the DB has been created. It's a bit the chicken and egg problem, but in any case, it's just a warning; odoo will operate normally (apart from _not_ using unaccent).

For these reasons, I close the PR. If you manage to provide a real fix, please feel free to reopen. :blush:

ap-wtioit:
@Yajo sorry if i was unclear:
The current problem is that config for unaccent is set to True at build time. Even if you set unaccent to False in Environment this only affects the database not the odoo.conf. Which causes warnings from Odoo at runtime because its missing the unaccent function, because database and odoo config are not matching for unaccent false.

Yajo:
You should update the scaffolding then. Build arg `CONFIG_BUILD` should be `false` to avoid generating conf at build time. Then, it is generated in entrypoint.

PR #143, opened 2018-05-21
[DCK] Add alternative domain redirection

Yajo:
https://github.com/containous/traefik/issues/1380 was fixed, it's time to support it.

### DIFF
--- a/.env
+++ b/.env
@@ -13,6 +13,7 @@ ODOO_PROXY_MODE=true
 # Domains
 DOMAIN_TEST=test.example.com
 DOMAIN_PROD=www.example.com
+DOMAIN_PROD_ALT=example.com
 
 # Database
 DB_VERSION=9.6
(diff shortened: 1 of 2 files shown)


.env
```diff
@@ -13,6 +13,7 @@ ODOO_PROXY_MODE=true
 # Domains
 DOMAIN_TEST=test.example.com
 DOMAIN_PROD=www.example.com
+DOMAIN_PROD_ALT=example.com
```
rafaelbn: Please could you add a comment? Like:

# You can add as many ALT domains as you want adding DOMAIN_PROD_ALT_2=subdomain.example.com, DOMAIN_PROD_ALT_3=othersubdomain.example.com, ...

Yajo: I only add comments with basic instructions in the scaffolding. Advanced instructions go in the project's README. I can add it there if you want, although it's pretty obvious IMHO :roll_eyes:

PR #146, opened 2018-06-04
Include pg_activity in base image

Yajo:
We find ourselves to be installing this quite often to track performance and concurrency issues, so it's better to bundle it.

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -121,7 +121,7 @@ RUN curl -SLo wkhtmltox.tar.xz https://github.com/wkhtmltopdf/wkhtmltopdf/releas
 # Other facilities
 WORKDIR /opt/odoo
 RUN pip install \
-    astor git-aggregator openupgradelib ptvsd==3.0.0 pudb wdb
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Why <1.5?

Yajo:
1.4.0 is the latest available. I don't think they will change dependencies before 1.5, if ever.

Recently we had https://github.com/Tecnativa/docker-odoo-base/pull/140 problem and I don't want to repeat it. Almost all packages are version-pinned here, so this one too.

pedrobaeza:
I prefer to not put that kind of limitation. If there's any incompatibility, then we will fix it freezing in that moment, but not before. And this kind of dependency is very soft, so in the worst case, we won't get pg_activity working, not Odoo.

Yajo:
Chances that something is broken are low, but are still higher than the chance that there's something new in pg_activity 1.5+ that we really can't wait for, so I'll stick with the conservationist preference this time if you don't mind.

pedrobaeza:
Sorry, but no, please free the version.

PR #148, opened 2018-06-19
Have vim available in the containers

chienandalu:
It'd be nice to have it :slightly_smiling_face:


pedrobaeza:
Why not use nano?

Issue #151, opened 2018-07-05
Some notes from initial use


gdgellatly:
I just started playing with this and I like it.  There are a few things I'd change in the scaffolding/conf.

1. dbtemplate in conf and set to template1 (or else create a template as variable). I notice Odoo has recently switched back to template0 default, but for me, if you want to populate extensions cleanly, template1 is better as long as you set locale.

2. Installing unaccent (or any other extension) - install on template1 (or above template var). The problem is otherwise first run the db doesn't exist.  I actually install on both template1 and attempt on the db, just in case.

3. db init args in common.yaml db-> environment IMHO POSTGRES_INITDB_ARGS: "--data-checksums --locale=C --encoding=UTF8" should be set as a minimum default. --data-checksums are critical for production and the performance hit is less than 1% for other environs.  Again the locale setting is critical, otherwise btree indexes useless for like and ilike searches.

gdgellatly:
Will do.

Yajo:
Closing due to repo split explained in #157. Feel free to reopen in https://github.com/Tecnativa/doodba-scaffolding if this is still an issue. Thanks!

PR #152, opened 2018-07-05
Set initdb args for Odoo in scaffolding common.yaml 

gdgellatly:
to ensure data integrity and index performance.

Address point 3 of #151

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -27,6 +27,7 @@ services:
         environment:
             POSTGRES_USER: "$DB_USER"
             POSTGRES_PASSWORD: "$DB_PASSWORD"
+            POSTGRES_INITDB_ARGS: "--data-checksums --locale=C --encoding=UTF8"
         volumes:
             - db:/var/lib/postgresql/data:z
 


pedrobaeza:
What about the LC_COLLATE thing?

gdgellatly:
? locale implicitly sets LC_COLLATE afaik as well as LC_TYPE

common.yaml
```diff
@@ -27,6 +27,7 @@ services:
         environment:
             POSTGRES_USER: "$DB_USER"
             POSTGRES_PASSWORD: "$DB_PASSWORD"
+            POSTGRES_INITDB_ARGS: "--data-checksums --locale=C --encoding=UTF8"
```
Yajo: setting `--locale=C` will affect users with other locales?

gdgellatly: Yes, it will make their databases perform properly.  The problem you usually have with alpine postgres is it uses en_US.utf8, which for Odoo breaks btree for text search.

In any case, from v12, I see Odoo is forcing C collation.  I could change it to just LC_COLLATE if that is what's wanted.

gdgellatly: I just read the docs and its hard to see the difference between setting locale and LC_COLLATE.  https://www.postgresql.org/docs/9.6/static/locale.html

if locale is set it sets LC_COLLATE.  If LC_COLLATE is set, it sets locale, which will then set LC_CTYPE and other LC vars unless specified.

Any specific setting for the other LC settings can be overidden in INITDB_ARGS.

gdgellatly: Probably the only way of doing in container would also solve the unaccent issue.

If database does not exist, CREATE DATABASE WITH LC_COLLATE etc.  Then you can add extensions.  Before all the auto install stuff of modules.  Picking LC_COLLATE is a one-shot thing for a new db.

idk to be honest, I just ported those 3 initdb args from my standard postgres compose file, but left out the non essential ones.

Yajo: I wouldn't like to create the database here manually.

Odoo, from v9+, automatically creates it. IMHO if we have to fix the DB creation process somewhere, it's there, directly upstream.

PR #153, opened 2018-07-05
Allow to specify a database template and install extensions there

gdgellatly:
if odoo database is not yet created.


gdgellatly:
Thats fine then we close the PR.  You can't install to template0 and you can't install to a non existent database.

gdgellatly:
Closing as installing to a template best left to the db not application image as we don't know what else db is used for.

PR #160, opened 2018-08-03
Add autoupdate timeout and document it

Yajo:
This feature is useful to find stale autoupdate jobs, which can randomly happen in parallel autoupdates.


11.0.Dockerfile
```diff
@@ -10,7 +10,8 @@ EXPOSE 8069 8072
 
 ARG WKHTMLTOPDF_VERSION=0.12.5
 ARG WKHTMLTOPDF_CHECKSUM='1140b0ab02aa6e17346af2f14ed0de807376de475ba90e1db3975f112fbd20bb'
-ENV DB_FILTER=.* \
+ENV AUTOUPDATE_TIMEOUT=30 \
```
pedrobaeza: There are certain processes that can take more than this timeout, like hooks, compute, and so on. This shouldn't been unconditional. As said in the task, you have to detect what is the last line to judge if it's a possible hang

Yajo: Yep, but it's also possible that we get hangs when the autoupdate is doing something different than updating table layouts, so it's better to start the script dumber and add specific cases when they actually happen. "Early optimization is the root of all performance problems", they say.

I'll raise this to 60 by default OTOH.

pedrobaeza: Please don't do that. We have a lot of processes that "hangs" the system more than 60 seconds. If you don't implement the constraint, don't put any timeout.

Yajo:
Travis :x: should get fixed after merging #159.

11.0.Dockerfile
Yajo: We are not talking about hanging the system, but about not producing any log before 60 seconds. Also, we are not talking about long install or migration processes, but about autoupdate only. The usage surface of this script is not that big. We need the timeout or the failure cannot be safely detected. You can always run the task with `-e AUTOUPDATE_TIMEOUT=9999999` if you don't trust it.

pedrobaeza: A computed field calculation doesn't produce any log until it's done, so yes, we are preventing this, and I want a sane default for this, so this one isn't. 600 is.

Yajo: 600 seconds means 10 min of production downtime. That's not a good default and violates the main point of this PR: unattended automated upgrades in an HA system, with a sane default for 99% of the times, while letting you handle those corner cases manually, if they ever happen.

1 min of downtime is still too much, that's why I put 30s on the first iteration, but OK, let's leave 1min. 10m definitely will not happen, unless the problem happens and there's a undeniable prove that's absolutely needed.

pedrobaeza: I don't want this. Let it without timeout if you are not willing to put that default, but are not going to assume a halted update because of this. I will accept it (even 30 seconds) if you implement the check on the last line being the one expected to hang ("Creating or updating..."), but not unconditionally.

Yajo: Feel free to open the pull request.

pedrobaeza: That's not the way to do it. You are the responsible of the task, so please do it as I'm telling you.

Yajo:
Setting as WIP. I'm gonna try to find a way to know if there are actually locked queries, instead of sniffing the logs, because if the blocking happens outside the controlled step, such as a pre-migration script or a data file loading, we're out of luck.

PR #166, opened 2018-09-11
Add missing Odoo 8.0 dependencies

crimoniv:
This PR adds a few missing dependencies for Odoo 8.0 instances. They're needed to generate the  **Workflow report** and enable document indexing with `document`.


crimoniv:
Thanks @Yajo!

Issue #19, opened 2017-03-01
[9.0] Werkzeug logs display the https proxy IP instead of the real IP


Yajo:
Using the 9.0 scaffolding, we get the proxy IP in the logs instead of the real one.

Seems related to having a version prior to https://github.com/pallets/werkzeug/pull/672 (0.11), but updating werkzeug to that version did not work.

It needs some further research.

Yajo:
I just checked that https://github.com/odoo/odoo/issues/18052#issuecomment-364703091 is true and using `werkzeug==0.14.1` simply works.

PR #168, opened 2018-09-19
Prepare for upcoming Odoo 12.0

Yajo:
Let's see how does this work.


12.0.Dockerfile
```diff
@@ -0,0 +1,165 @@
+FROM python:3.7-stretch AS base
```
pedrobaeza: This should be 3.5 for adjusting to Odoo. There are know incompatibilities with this latest version.l

12.0.Dockerfile
```diff
@@ -0,0 +1,165 @@
...
+    WDB_SOCKET_SERVER=wdb \
+    WDB_WEB_PORT=1984 \
+    WDB_WEB_SERVER=localhost
+
+# Other requirements and recommendations to run Odoo
+# See https://github.com/$ODOO_SOURCE/blob/$ODOO_VERSION/debian/control
+RUN apt-get -qq update \
+    && apt-get -yqq upgrade \
+    && apt-get install -yqq --no-install-recommends \
+        chromium \
```
pedrobaeza: I have needed saasc for making it work

12.0.Dockerfile
```diff
@@ -0,0 +1,165 @@
...
+    && ln /usr/local/bin/direxec common/entrypoint \
+    && ln /usr/local/bin/direxec common/build \
+    && chmod -R a+rx common/entrypoint* common/build* /usr/local/bin \
+    && chmod -R a+rX /usr/local/lib/python3.7/dist-packages/odoobaselib \
+    && sync
+
+# Execute installation script by Odoo version
+# This is at the end to benefit from cache at build time
+# https://docs.docker.com/engine/reference/builder/#/impact-on-build-caching
+ARG ODOO_SOURCE=OCA/OCB
```
pedrobaeza: Use odoo/odoo and saas-11.5 for now

12.0.Dockerfile
```diff
@@ -0,0 +1,165 @@
...
+ARG ODOO_VERSION=12.0
+ENV ODOO_VERSION="$ODOO_VERSION"
+RUN pip install -r https://raw.githubusercontent.com/odoo/odoo/saas-11.5/requirements.txt \
+    && pip install pg_activity \
+    && (python3 -m compileall -q /usr/local/lib/python3.7/ || true)
+
+# HACK Special case for Werkzeug
+USER odoo
+RUN pip install --user Werkzeug==0.14.1
```
pedrobaeza: Why freezing this version? Use the one declared in Odoo requirements.txt

12.0.Dockerfile
Yajo: AFAIK from https://github.com/odoo/odoo/pull/26587 odoo 11 is compatible already. Are you sure master is not?

12.0.Dockerfile
Yajo: I think it's not needed if you install the python sass library. I want to test

12.0.Dockerfile
Yajo: See https://github.com/Tecnativa/doodba/issues/19

12.0.Dockerfile
Yajo: It's done below, I just wanted to leave this as it is for a week, it shouldn't harm much.

12.0.Dockerfile
pedrobaeza: I prefer to stick to exact compatible version in this case. No real advantages of using a later version, as we can't use new idiomatics. Is there any performance advantage? And why use pinned module versions, but not pin this?

Yajo: Pinning versions gives stability, while upgrading usually gives new features and better performance. That's why, as a general criteria, it's a good idea to upgrade dependencies before doing a new release, and freeze them once done.

However, I don't have benchmarks of Odoo on python 3.7 vs 3.5, if that's your question.

BTW, you didn't answer my question: are you sure odoo 12 is not compatible with python 3.7?

pedrobaeza: > are you sure odoo 12 is not compatible with python 3.7?

No, but Odoo neither. My point is that we shouldn't risk for not known advantages. Even with advantages, we have a serious drawback: we will be able to develop in our stack things that are incompatible with the rest of Odoo because we don't remember that are 3.6 or 3.7 features, and as it works in our development environment...

Yajo: OK

PR #171, opened 2018-09-27
Use OCB 12.0

Yajo:
- Branch 12.0 of OCB has been released, so use it now.
- Enable some tests that should work now. There are still anothers that are disabled.
- Use Python 3.7, now that it has been discovered [it is officially supported][1] ~~and [is going to be tested in OCA repos][2]~~.

[1]: https://twitter.com/__yajo/status/1045228965928062977
[2]: https://twitter.com/moylop260/status/1045230320403062784

### DIFF
--- a/12.0.Dockerfile
+++ b/12.0.Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.5-stretch AS base
+FROM python:3.7-stretch AS base
 
 # Enable Odoo user and filestore
 RUN useradd -md /home/odoo -s /bin/false odoo \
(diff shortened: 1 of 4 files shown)


12.0.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.5-stretch AS base
+FROM python:3.7-stretch AS base
```
pedrobaeza: :-1: to this change. We have discussed already about this.

Yajo: Yes, I thought about [what you said](https://github.com/Tecnativa/doodba/pull/168#discussion_r219116920):

> we will be able to develop in our stack things that are incompatible with the rest of Odoo because we don't remember that are 3.6 or 3.7 features

I think you didn't realize the opposite problem is also true: other people can do exactly the same, and we'd be affected because we are the ones using an older Python version.

If I have to choose among being affected by a problem or affecting others, I think the sanest approach, sadly, is affecting others.

PR #176, opened 2018-10-19
Install compass

Yajo:
Odoo 12.0 core uses compass:

- https://github.com/odoo/odoo/blob/cb8dcb5913de3d3067e4ba1ad218dc91c9ef716d/odoo/addons/base/models/assetsbundle.py#L571
- https://github.com/odoo/odoo/blob/1868713dbd07e0b518f91dffe73e62d85e6ab9a6/odoo/addons/base/static/src/css/description.sass#L2-L3

Let's bundle it in the image, by default.

### DIFF
--- a/12.0.Dockerfile
+++ b/12.0.Dockerfile
@@ -43,7 +43,7 @@ RUN apt-get -qq update \
         gnupg2 \
         locales-all \
         nano \
-        ruby \
+        ruby-compass \
         telnet \
         vim \
         zlibc \


Yajo:
It seems like Odoo is confusing here because it doesn't actually use ruby sass for its bundled .scss files, just for .sass files. In practice, this means that sass and scss files are not compatible between themselves, so the compass support would only work for .sass files, not for .scss. This is kind of weird, but I think I'm not gonna merge this for the time being, unless I get more time and insight in the matter.

If you need things from compass, which only works with ruby sass, probably it's better that you use https://www.bourbon.io/ if possible, which adds little dependency and works fine, no matter the underlying implementation.

PR #177, opened 2018-11-05
Allow to find dependant addons only

Yajo:
This tool is required to be able to apply the same testing workflow as found in OCA, namely:

1. Install dependencies of chosen addons.
2. Test & install chosen addons.

See https://github.com/OCA/server-ux/pull/32 and https://github.com/odoo/odoo/pull/27883 to understand the motivation behind this. Take a :coffee: first :stuck_out_tongue_winking_eye:.

### DIFF
--- a/bin/addons
+++ b/bin/addons
@@ -24,6 +24,9 @@ parser.add_argument(
     "-c", "--core", action="store_true",
     help="Use all Odoo core addons")
 parser.add_argument(
+    "-d", "--dependencies", action="store_true",
+    help="Use only dependencies of selected addons")
+parser.add_argument(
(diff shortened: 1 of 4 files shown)


PCatinean:
@Yajo If i'm not mistaken this enhancement might have introduced a bug that I'm experiencing.

When listing private addons with -i flag it also picks up README.md, oca_requirements.txt and files like that. 

Scripts from the qa image like flake8 for example fail as they parse also these files

Yajo:
Hmm I haven't seen any of those problems... :thinking: 

Please open a new issue and provide:

- Relevant `repos.yaml`
- Relevant `addons.yaml`
- Logs

Thanks!

PR #178, opened 2018-11-05
Update python in v11 image (switch base)

Yajo:
The outdated and buggy Python 3.5 version that Debian 9 ships was triggering werid behaviors in some production instances.

I switch here the base to Docker Hub's `python:3.5-stretch`, which shares the same underlying OS, but uses latest Python 3.5 (3.5.6 as of today) and gets updates faster.

The main upgrade problem would be that projects adding `python-dev` to their `apt_build.txt` files should remove it now. Not a big problem nor fix, so it seems legit for the stable release.

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:9 AS base
+FROM python:3.5-stretch AS base
 
 # Enable Odoo user and filestore
 RUN useradd -md /home/odoo -s /bin/false odoo \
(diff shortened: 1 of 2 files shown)


pedrobaeza:
And if you need that library for installing some pip libraries?

Yajo:
It's shipped now in the base image, so no need to install it. In fact, if you install it, you'd end up having 2 different versions of python development headers, so to avoid confusion, just remove that line from dependencies and you're done.

pedrobaeza:
OK, good to know. Maybe you should put an announcement in the breaking changes thread.

Yajo:
I'm writing it right now :wink:

Issue #189, opened 2018-11-20
400-clean cannot remove link as subdirectory


remi-filament:
During build, while trying to clean out unused modules, I face the following error : 
```
INFO:docker-odoo-base:Removing subdirectory /opt/odoo/custom/src/bank-statement-import/account_bank_statement_import_camt_oca
Traceback (most recent call last):
  File "400-clean", line 41, in <module>
    shutil.rmtree(subfull)
  File "/usr/lib/python2.7/shutil.py", line 232, in rmtree
    onerror(os.path.islink, path, sys.exc_info())
  File "/usr/lib/python2.7/shutil.py", line 230, in rmtree
    raise OSError("Cannot call rmtree on a symbolic link")
OSError: Cannot call rmtree on a symbolic link
```
The issue comes from a specific module in bank-statement-import OCA repo : [account_bank_statement_import_oca](https://github.com/OCA/bank-statement-import/blob/10.0/account_bank_statement_import_camt_oca) which is a symlink to the same module without _oca at the end.

Not cleaning the repositories works around the issue (adding CLEAN = false), but we may face the same issue with another module,

Since 400-clean is already not cleaning individual files at root of directories and subdirectories, a possible solution would be to check if directory or subdirectory are not links as well (although it would then leave dead links, therefore not sure this is ideal solution).
If you consider this would do the trick, I can provide a PR with the following code changes:

L22 : `if not (os.path.isdir(full) or os.path.islink(full)):`
L35 : `if not (os.path.isdir(subfull) or os.path.islink(subfull)):`

Yajo:
Please open that PR and let's discuss on it.

Maybe you can try an easier approach of something similar to:

```python
try:
  shutil.rmtree(subfull)
except OSError:
  os.unlink(subfull)
```

[EAFP](https://docs.python.org/3/glossary.html#term-eafp).

remi-filament:
Done, thanks for the tip with try/catch solution, it worked.

PR #190, opened 2018-11-21
Fix removal of linked directories

remi-filament:
This PR is proposed to fix the issue #189 

As proposed in the mentioned issue, a try/catch has been implemented and successfully tested for my use case, however, it may be that OSError is raised for another reason than because directory is a link and therefore the script would still fail when trying to unlink it...

### DIFF
--- a/build.d/400-clean
+++ b/build.d/400-clean
@@ -25,7 +25,10 @@ for directory in os.listdir(SRC_DIR):
     # Remove directories not listed in addons.yaml
     if directory not in repos:
         logger.info("Removing directory %s", full)
-        shutil.rmtree(full)
+        try:
+          shutil.rmtree(full)
+        except OSError:
+          os.unlink(full)
         continue
 
     # Traverse addons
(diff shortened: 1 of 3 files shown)


remi-filament:
That was mostly hypothetical, for instance in case full or subfull does not exist. Reviewing the code, you should not reach that point if full  / subfull does not exist (except if directories were to be removed externally at the same time as the script is running, but this should not happen), so I guess this is safe to go ahead with proposed change.

Yajo:
No, sorry, I meant a test case for the problem you are actually fixing: the non removal of a symlink that should be removed.

remi-filament:
I am not sure if you mean describe the test case.

The test case would then be the following : 
1. Add bank_statement_import from OCA in repos.yaml : 
```
./bank-statement-import:
    defaults:
        depth: $DEPTH_DEFAULT
    remotes:
        oca: https://github.com/OCA/bank-statement-import.git
    target:
        oca $ODOO_VERSION
    merges:
        - oca $ODOO_VERSION
```

2. Add module account_bank_statement_import_ofx in addons.yaml:
```
bank-statement-import:
    -  account_bank_statement_import_ofx
```
3. Build docker with the above files and environment variable CLEAN set to true.

**Expected result** : all directories but account_bank_statement_import_ofx are removed from /opt/odoo/custom/src/bank_statement_import

PR #192, opened 2018-12-18
Bundle QA dependencies

Yajo:
Following the QA redesign stated in https://github.com/Tecnativa/doodba-qa/issues/8, here I'm bundling all QA dependencies in a separate virtualenv located in `/qa` folder, inside the image, to make QA easier to execute under any environment.


pedrobaeza:
Why don't install requirements.txt from MQT?

Yajo:
Because only flake8 is used from MQT, so no need to install all other requirements that are not applicable for Doodba.

pedrobaeza:
I'm seeing all of this installed:

click
coverage
flake8
pylint-odoo
six

which are also on MQT

pedrobaeza:
That's very weird that MQT doesn't include proper libraries... but that's another story, so I don't insist more

Yajo:
Lol, yes! But in any case we only use a couple of pieces from there, so it's not our problem here :smile:

PR #194, opened 2019-01-07
Use backwards-compatible universal_newlines argument

Yajo:
I was using the `text=True` argument, only in PY3+, for `subprocess.check_call`, but it turns out that argument only exists in PY3.7+, and this code is executed also by PY2.7 and PY3.5, depending on the Odoo image version.

Instead, I just use now `universal_newlines=True`, which is the backwards-compatible equivalent to `text=True`. We'll change that when that backwards compatibility is removed and/or we only support PY3.7+ (a.k.a. only Odoo 12.0+).

This fixes the red build introduced by #193.

### DIFF
--- a/qa/insider
+++ b/qa/insider
@@ -42,7 +42,8 @@ try:
     subprocess.check_call(
         cmd,
         stdout=sys.stderr,
-        **({"text": True} if sys.version_info.major > 2 else {}),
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Only a note: Odoo v13 will be Python 3.6+ compatible, not 3.7+

Issue #67, opened 2017-08-07
📌⚠️  Possible retrocompatibility breaking changes announcements (SUBSCRIBE if you use Doodba)


Yajo:
This project is currently in a production-ready but evolving status, so some not-retrocompatible changes might come in.

This issue will be a retrocompatibility breakage tracker. Please subscribe here if you use this project in production, because you will care of these changes.

Any announcement will include instructions on how to pin the image versions that supported old behaviour, and upgrade instructions.

Yajo:
# New backup system

When #68 gets merged, we will start using https://github.com/Tecnativa/docker-duplicity project to generate backups.
## Why will this happen?

Duplicity is better docummented, supported, efficient, supports many backends (although default will be S3 as usual)...

## What does it break?

New backup system is not compatible with old one. It uses different configuration, different storage, different everything.

## Last non-breaking image/scaffolding version

- Scaffolding: https://github.com/Tecnativa/docker-odoo-base/commit/fc8767955218be6e152b5dd78a3203e1b7863d9c
- Image not affected by this change.

## How to update?

Configure your S3 backend to move objects prefixed with `archive-` into glacier after 1 month, and to expire everything after 3 months.

If you need to restore an old backup, you will have to roll back your project to the last commit before updating the scaffolding, and use the old restore process.

Yajo:
# Removing support for `/opt/odoo/custom/src/requirements.txt` file

Support was deprecated after #33. Now it's going to be removed.

## When will this happen?

When #28 gets merged. Subscribe there for updates.

## Why will this happen?

Deprecated since 2 months ago, it's time to remove it... We have a better location for it now.

## What does it break?

Builds using that file will not get pip dependencies installed.

## Last non-breaking image/scaffolding version

- Scaffolding: https://github.com/Tecnativa/docker-odoo-base/commit/1b0d4a6d05904bf6f11ea2aef319bf7c8d09a5ef
- Images:
  - 8.0: tecnativa/odoo-base@sha256:736196b6048bfba231092e5d4ddbb6edcf0838b1daaa4c6b6a5b3b17f47ad270
  - 9.0: tecnativa/odoo-base@sha256:13e9cf5a259eba70d559013a854a3cceb40d36c95ff3f50a83670385b71b078a
  - 10.0: tecnativa/odoo-base@sha256:3fb04373e16b25081d8d571b1faf58ae57902611b8a740004ae489356f9c924b

## How to update?

[...]

PR #199, opened 2019-02-05
Downgrade Python version to 3.5 in v12

Yajo:
This is the minimal version that is supported.


Yajo:
Yes, I just would feel better if you hit the merge button. :point_right: :point_left:  You know I don't really like this change. :roll_eyes: 

Just let me notify it in #67.

Yajo:
Ready

PR #203, opened 2019-02-14
Allow to set list_db parameter by env variable

Yajo:
This should allow us to disable the database manager at will in environments, removing the need to protect those controllers through the external proxy.


patovala:
How is this related to copier template? cause in the questions/answers I specified that I want to list_db.  I was lost trying to show databases until I see your PR. The LIST_DB env should be documented somewhere else, or I'm missing something.

Yajo:
Please use the appropriate channels of communication. This thread ended 3 years ago... If you have any questions, open a new discussion in the forum.

PR #210, opened 2019-03-22
[IMP] Filter out especific addons from a repository

lreficent:
The intent is to be able to exclude desired addons from a repository using this syntax:

```yml
some-repo:
  - "*"
  - "!addon_to_exclude"
```

@Yajo

### DIFF
--- a/lib/doodbalib/__init__.py
+++ b/lib/doodbalib/__init__.py
@@ -113,6 +113,8 @@ def addons_config(filtered=True, strict=False):
     logger.debug("Merged addons definition before expanding: %r", all_globs)
     # Expand all globs and store config
     for repo, partial_globs in all_globs.items():
+        to_exclude = [x[1:] for x in
+                      filter(lambda g: g.startswith('!'), partial_globs)]
         for partial_glob in partial_globs:
             logger.debug("Expanding in repo %s glob %s", repo, partial_glob)
             full_glob = os.path.join(SRC_DIR, repo, partial_glob)
@@ -134,6 +136,9 @@ def addons_config(filtered=True, strict=False):
                     logger.debug(
                         "Skipping '%s' as it is not a valid Odoo "
                         "module", addon)
+                    continue
+                if addon in to_exclude:
+                    logger.debug("Excluded '%s'", addon)
                     continue
                 logger.debug("Registering addon %s", addon)
                 addon = os.path.basename(addon)
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -104,8 +104,8 @@ class ScaffoldingCase(unittest.TestCase):
                 # absent_addon is missing and should fail
                 ("bash", "-c", "! addons list -px"),
                 # hr_payroll should have been filtered out
-                ("test", "!", "-d", "auto/addons/hr_payroll"),
-                # ("test", "-d", "auto/addons/hr"),
+                # ("test", "!", "-d", "auto/addons/hr_payroll"),
+                ("test", "-e", "auto/addons/hr"),
                 # ("test", "-d", "auto/addons/hr_payroll_account"),
             )
             self.compose_test(
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -104,7 +104,7 @@ class ScaffoldingCase(unittest.TestCase):
                 # absent_addon is missing and should fail
                 ("bash", "-c", "! addons list -px"),
                 # hr_payroll should have been filtered out
-                ("bash", "-c", "test ! -d auto/addons/hr_payroll"),
+                # ("test", "! -d", "auto/addons/hr_payroll"),
                 ("test", "-d", "auto/addons/hr"),
                 ("test", "-d", "auto/addons/hr_payroll_account"),
             )
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -105,7 +105,7 @@ class ScaffoldingCase(unittest.TestCase):
                 ("bash", "-c", "! addons list -px"),
                 # hr_payroll should have been filtered out
                 # ("test", "!", "-d", "auto/addons/hr_payroll"),
-                ("test", "-e", "auto/addons/hr"),
+                # ("test", "-e", "auto/addons/hr"),
                 # ("test", "-d", "auto/addons/hr_payroll_account"),
             )
             self.compose_test(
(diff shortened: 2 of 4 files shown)


tests/scaffoldings/dotd/custom/src/addons.yaml
```diff
@@ -3,6 +3,10 @@ other-doodba/odoo/src/private:
   - dummy_addon
   - product
 
+odoo/addons:
+  - *
+  - !hr_payroll
```
Yajo: The `!` character has a special meaning for YAML (it's breaking Travis here). Can't you use a different one? Otherwise, enclose it in quotes.

Also, as commented above, please append a `*`, so we can actually test that globs work as expected.

tests/__init__.py
```diff
@@ -103,6 +103,10 @@ def test_addons_filtered(self):
                 ("bash", "-c", 'addons list -c | grep ,crm,'),
                 # absent_addon is missing and should fail
                 ("bash", "-c", "! addons list -px"),
+                # hr_payroll should have been filtered out
+                ("bash", "-c", "test ! -d auto/addons/hr_payroll"),
+                ("test", "-d", "auto/addons/hr"),
+                ("test", "-d", "auto/addons/hr_payroll_account"),
```
Yajo: Please use a glob in the test and inverse this test (adding `"!", `)

tests/__init__.py
```diff
@@ -103,6 +103,10 @@ def test_addons_filtered(self):
                 ("bash", "-c", 'addons list -c | grep ,crm,'),
                 # absent_addon is missing and should fail
                 ("bash", "-c", "! addons list -px"),
+                # hr_payroll should have been filtered out
+                ("bash", "-c", "test ! -d auto/addons/hr_payroll"),
```
Yajo: No need for bash here.

lib/doodbalib/__init__.py
```diff
@@ -113,6 +113,8 @@ def addons_config(filtered=True, strict=False):
     logger.debug("Merged addons definition before expanding: %r", all_globs)
     # Expand all globs and store config
     for repo, partial_globs in all_globs.items():
+        to_exclude = [x[1:] for x in
+                      filter(lambda g: g.startswith('!'), partial_globs)]
```
Yajo: It seems to me like this will not work with glob patterns. :thinking:  Please test that use case (see comments below) and make it pass. :blush:

tests/scaffoldings/dotd/custom/src/addons.yaml
lreficent: ups, yes, of course :sweat_smile: 

Let's do this one first and see the effect in travis

tests/__init__.py
lreficent: @Yajo how would it be? `("test", "! -d", "auto/addons/hr_payroll")`?

lib/doodbalib/__init__.py
lreficent: this is using the strings that were feched before from the addons.yml file and extracting the lines starting with `!` to use it afterwards, it does not modify the result of `found = glob(full_glob)`

so it is not expecting the result of the glob to change but to store in a list the modules to exclude later simple with this:
```python
                if addon in to_exclude:
                    logger.debug("Excluded '%s'", addon)
                    continue
```

So it is not relying in the glob functionalities but doing a work-around.

tests/__init__.py
lreficent: what do you mean, I added three tests (don't know if the syntax is correct though) one to test that hr_payroll is filtered out and 2 more for testing that hr and hr_payroll_account still remaing there. Becasuse if you use the valid glob syntax `- "[!hr_payroll]" it will filter out all of them.

tests/__init__.py
Yajo: `("test", "!", "-d", "auto/addons/hr_payroll")`

lib/doodbalib/__init__.py
Yajo: I think it would be better to do something like this pseudo-code:

```python
matches = {"include": set(), "exclude": set()}
for glob in globs:
  # Know if the glob is to include or exclude
  mode = "exclude" if glob.startswith("!") else "include"
  if mode == "exclude":
    # Remove "!" from the start
    glob = glob[1:]
  matched_dirs = expand_glob(glob)
  matches[mode] |= matched_dirs
# The final list of addons to include removes the excluded ones
winners = matches["include"] - matches["exclude"]
```

tests/scaffoldings/dotd/custom/src/addons.yaml
```diff
@@ -3,6 +3,10 @@ other-doodba/odoo/src/private:
   - dummy_addon
   - product
 
+odoo/addons:
+  - "*"
+  - "!hr_payroll"
```
Yajo: Answering here to https://github.com/Tecnativa/doodba/pull/210/files#r269939304, I mean that you please add a negative glob here.

If users can use globs to include addons, it is expected that they can use it also to exclude. The code you pushed doesn't support exclusion globs, nor does it know what to do if there are only negative globs, so that's why I ask you to add a test with globs and make it pass (to have [TDD](https://en.wikipedia.org/wiki/Test-driven_development)).

A good example:

```yaml
odoo/addons:
  - "!hr_payroll*"
```

Then, test that:

- i.e. `sale` addon is present (because of the implicit `"*"` (only implicit in the special `odoo/addons` and `private` sections)).
- `hr_payroll` and `hr_payroll_account` are absent (because it matches the negative glob).

tests/__init__.py
lreficent: Thanks, thought I could only pass three elements :sweat_smile:

tests/scaffoldings/dotd/custom/src/addons.yaml
lreficent: OK, I understand you now, you was thinking on use negative globs as exclusion mechanism. Will try!

amon-ra:
Maybe adding regular expressions instead globs are simpler and more powerful

Yajo:
I prefer not. They are more powerful, but not simpler. Also it would break existing deployments.

Yajo:
This became stale, closing. Please reopen if needed. Thanks!

PR #212, opened 2019-03-26
Disable the jessie-updates repo in old images

Yajo:
Images for Odoo v8-10, based on Debian 8, started to display errors when doing `apt-get update`.

The reason is that [the `jessie-updates` apt repo has been removed from apt mirrors](https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html).

I might consider updating the base OS, but for now IMHO the safest thing to do is to remove that repo.

### DIFF
--- a/8.0.Dockerfile
+++ b/8.0.Dockerfile
@@ -37,7 +37,8 @@ ENV DB_FILTER=.* \
 
 # Other requirements and recommendations to run Odoo
 # See https://github.com/$ODOO_SOURCE/blob/$ODOO_VERSION/debian/control
-RUN apt-get update \
+RUN sed -Ei 's@(^deb http://deb.debian.org/debian jessie-updates main$)@#\1@' /etc/apt/sources.list \
+    && apt-get update \
     && apt-get -y upgrade \
     && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But this means kernels and other stuff without security updates, isn't it?

Yajo:
Well... yes and no.

- The kernel used by containers is the host's kernel, so bugfixes land there.
- As a good practice, we never do an `apt-get upgrade` when building the image, because in case there's a security patch, the base image is rebuilt at the hub, and that one in turn triggers a rebuild of ours because they are configured like that.
- Rebasing images in debian 9 could potentially break many subimages with dependencies downstream.
- I'm not disabling the line `deb http://security.debian.org/debian-security jessie/updates main`, which I think is the one that should provide the security updates (apart from the comments above).
- Most important: after all, today odoo 8 and 9 have no security maintenance either, nor will have odoo 10 after next october, nor will have python 2 itself after next year. So projects should upgrade to odoo 11+ ASAP, but that's not Debian's fault, it's just normal software lifecycle.

So I think this change is the MVP to keep on going.

pedrobaeza:
OK then

PR #218, opened 2019-04-16
Include ffmpeg in v12

Yajo:
Odoo v12 supports logging a screencast of failed tours when `ffmpeg` is present and `--logfile` is supplied.

Obviously, including `ffmpeg` here. Screencasts everywhere! 📹

### DIFF
--- a/12.0.Dockerfile
+++ b/12.0.Dockerfile
@@ -41,6 +41,7 @@ RUN apt-get -qq update \
     && apt-get -yqq upgrade \
     && apt-get install -yqq --no-install-recommends \
         chromium \
+        ffmpeg \
         fonts-liberation2 \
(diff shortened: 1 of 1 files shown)


pedrobaeza:
It would be great a mention of that in the README of Doodba

Yajo:
I think a better place is Doodba QA. This is standard Odoo 12 now, and as always Doodba includes all possible dependencies. However, the possibility to publish tour screenshots and screencasts as browsable artifacts in your CI is something more Doodba QA specific and it seems like a good place to document this. :thinking:

pedrobaeza:
OK, as you prefer. The idea is to have all this knowledge gathered for summarizing things than our docker environment can do.

PR #221, opened 2019-05-07
[IMP] enable using other uids than 1000 for odoo user

ap-wtioit:
also fix permissions for volume on startup

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -195,6 +195,13 @@ ONBUILD ENV ADMIN_PASSWORD="$ADMIN_PASSWORD" \
             WITHOUT_DEMO="$WITHOUT_DEMO"
 ONBUILD ARG LOCAL_CUSTOM_DIR=./custom
 ONBUILD COPY $LOCAL_CUSTOM_DIR /opt/odoo/custom
+
+# enable setting custom uids for odoo user during build of scaffolds
+ONBUILD ARG ODOO_UID=1000
+ONBUILD ARG ODOO_GID=1000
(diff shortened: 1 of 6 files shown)


11.0.Dockerfile
```diff
@@ -195,6 +195,13 @@ ONBUILD ENV ADMIN_PASSWORD="$ADMIN_PASSWORD" \
             WITHOUT_DEMO="$WITHOUT_DEMO"
 ONBUILD ARG LOCAL_CUSTOM_DIR=./custom
 ONBUILD COPY $LOCAL_CUSTOM_DIR /opt/odoo/custom
+
+# enable setting custom uids for odoo user during build of scaffolds
+ONBUILD ARG ODOO_UID=1000
+ONBUILD ARG ODOO_GID=1000
+ONBUILD ENV ODOO_UID="$ODOO_UID" \
```
Yajo: I don't think the ENV is needed. Instead, we could change [the `USER odoo` line below](https://github.com/Tecnativa/doodba/pull/221/files#diff-0ad4db09fb0257b087f8a07f47f5c354L205) to set the UID/GID of the `odoo` running user on creation. Possibly it would be the best, although possibly instructions in line 4 need to change too.

ap-wtioit: i think the env is needed for the entrypoint.d script to pass the values into the run (if we get rid of the entrypoint script then we can remove those env lines as well)

ap-wtioit:
was superseded by https://github.com/Tecnativa/doodba/pull/224

PR #227, opened 2019-07-08
Smarter, less strict, config generator

Yajo:
Before Odoo v11, Odoo wasn't so strict when reading configuration. Since v11, where Python 3 started to be used, it used the new default `strict=True` mode to read configuration files.

The strict mode forbids section and option repetitions. Although that seems legit for most apps (even Odoo itself), in Doodba we provide a set of predefined defaults that downstream scaffoldings should be able to override.

To allow such behavior, now the script that generates the final config will do it in `strict=False` mode, but will remove duplicates and just use the latest option defined, effectively producing a strict config file out of a collection of non-strict configs.

Of course, options until now are respected, like output of result while in `LOG_LEVEL=DEBUG` or env variable substitution.

Full power! 💪

### DIFF
--- a/bin/config-generate
+++ b/bin/config-generate
@@ -1,9 +1,51 @@
-#!/bin/bash
-# Generate Odoo server configuration from templates
-set -e
-src="/opt/odoo/common/conf.d/* /opt/odoo/custom/conf.d/*"
-log INFO Merging $(ls $src | wc -l) configuration files in $OPENERP_SERVER
-conf=$(cat $src | envsubst)
-log DEBUG "Resulting configuration:
-$conf"
-echo "$conf" > $OPENERP_SERVER
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""Generate Odoo server configuration from templates"""
+
+import os
+from contextlib import closing
+from doodbalib import logger
+from string import Template
+
+try:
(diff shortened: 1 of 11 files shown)


bin/config-generate
```diff
@@ -1,9 +1,40 @@
...
+
+try:
+    from configparser import RawConfigParser
+    parser = RawConfigParser(strict=False)
+except ImportError:
+    # Python 2, where strict=True doesn't exist
+    from ConfigParser import RawConfigParser
+    parser = RawConfigParser()
+
+TARGET_FILE = os.environ["OPENERP_SERVER"]
```
pedrobaeza: Isn't another variable in new versions (like `ODOO_SERVER`)? If so I think we should fallback to that one for the day that Odoo decides to remove the deprecated variable name for not having to make any change.

Yajo: They added `ODOO_RC` in v10. I was using this one because it's the one compatible with all supported versions, but let me add support for the other one, yes.

Yajo: Oh wait, the problem is that if I add support here, I'll have to check the current odoo version to know if I should support that new variable or not (wether it's in v8/v9 or it's newer). Is it worth it, knowing this one works always? :thinking:

PR #228, opened 2019-07-13
Fix build if GID exits

amon-ra:
This fixes docker build when try to asign a GID that already exists in parent image.

It is a very lite patch.

### DIFF
--- a/bin/config-generate
+++ b/bin/config-generate
@@ -32,7 +32,7 @@ CONFIG_FILES = []
 logger.info("Merging found configuration files in %s", TARGET_FILE)
 for dir_ in CONFIG_DIRS:
     try:
-        for file_ in os.listdir(dir_):
+        for file_ in sorted(os.listdir(dir_)):
             parser.read(os.path.join(dir_, file_))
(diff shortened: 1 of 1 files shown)


11.0.Dockerfile
```diff
@@ -188,8 +188,8 @@ ONBUILD ARG UID=1000
 ONBUILD ARG GID=1000
 
 # Enable Odoo user and filestore
-ONBUILD RUN groupadd -g $GID odoo \
-    && useradd -md /home/odoo -s /bin/false -u $UID -g $GID odoo \
+ONBUILD RUN groupadd -g $GID odoo || true
```
Yajo: Do this:

```suggestion
ONBUILD RUN (groupadd -g $GID odoo || true) \
```

... and let the next line untouched, please.

Also do the same for all other dockerfiles.

amon-ra: I have updated my branch.

amon-ra:
Does anyone know why is travis failing?

pedrobaeza:
@Yajo is on holidays, so we should wait until he's come back

PR #235, opened 2019-08-16
Support Odoo v7

Yajo:
This support is suboptimal, not targeted for production, just for migrations.


.travis.yml
```diff
@@ -20,10 +20,11 @@ env:
...
   matrix:
+    - DOCKER_TAG=7.0
     - DOCKER_TAG=8.0
-    - DOCKER_TAG=9.0
-    - DOCKER_TAG=10.0
-    - DOCKER_TAG=11.0
+    - DOCKER_TAG=9.0 PG_VERSIONS=9.6
+    - DOCKER_TAG=10.0 PG_VERSIONS=10
+    - DOCKER_TAG=11.0 PG_VERSIONS=11
     - DOCKER_TAG=12.0
```
pedrobaeza: And why not here putting `PG_VERSIONS=11`?

Yajo: It's the default. I just leave room for future versions.

Issue #247, opened 2019-10-07
Building doodba for version 13 fails since pull request got merged


ap-wtioit:
Because https://github.com/acsone/click-odoo/pull/34 got merged and the branch Tecnativa:parse-version deleted the build for version 13.0 fails now with:
```
Collecting git+https://github.com/Tecnativa/click-odoo.git@parse-version
  Cloning https://github.com/Tecnativa/click-odoo.git (to revision parse-version) to /tmp/pip-req-build-_j0ntm02
  Running command git clone -q https://github.com/Tecnativa/click-odoo.git /tmp/pip-req-build-_j0ntm02
  WARNING: Did not find branch or tag 'parse-version', assuming revision or ref.
  Running command git checkout -q parse-version
  error: pathspec 'parse-version' did not match any file(s) known to git
ERROR: Command errored out with exit status 1: git checkout -q parse-version Check the logs for full command output.
```
https://github.com/Tecnativa/doodba/commit/f97341b22563f1071c8ee3fd473048187fcb8324#r35384832

(i will try to switch to `pip install -U git+https://github.com/acsone/click-odoo.git@eb9e5568baead002da877c729b57bd9274d3f4ec`)

PR #248, opened 2019-10-07
fix build errors by using merged pull request for click-odoo

ap-wtioit:
previously a now deleted branch was used (parse-version was merged in https://github.com/acsone/click-odoo/commit/eb9e5568baead002da877c729b57bd9274d3f4ec)

fixes https://github.com/Tecnativa/doodba/issues/247

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -71,7 +71,7 @@ RUN gem install --no-rdoc --no-ri --no-update-sources bootstrap-sass --version '
 WORKDIR /opt/odoo
 RUN pip install \
         astor \
-        git+https://github.com/acsone/click-odoo-contrib.git@1a8ef8bcbeaf8ecbfdfe9b83936ad9f2dd819fc3 \
+        click-odoo-contrib \
         git-aggregator \
         pg_activity \
         ptvsd \
--- a/12.0.Dockerfile
+++ b/12.0.Dockerfile
@@ -67,7 +67,7 @@ RUN ln -s /usr/bin/nodejs /usr/local/bin/node \
 WORKDIR /opt/odoo
 RUN pip install \
         astor \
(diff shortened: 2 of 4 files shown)


ap-wtioit:
@pedrobaeza at a second look it would have been better to delete the line (click-odoo is a dependency of click-odoo-contrib and gets installed anyways (current version 1.4.2))

if we want to pin the version this pull request seems ok, but then i also would prefer to use a tagged version instead of the commit hash `git+https://github.com/acsone/click-odoo-contrib.git@1a8ef8bcbeaf8ecbfdfe9b83936ad9f2dd819fc3`  (`git+https://github.com/acsone/click-odoo-contrib.git@1.8.0`) https://github.com/acsone/click-odoo-contrib/commit/1a8ef8bcbeaf8ecbfdfe9b83936ad9f2dd819fc3

what do you think?

ap-wtioit:
Sorry for force pushing a lot.

if i use just git+https://github.com/acsone/click-odoo-contrib.git@1a8ef8bcbeaf8ecbfdfe9b83936ad9f2dd819fc3 in the 13.0.Dockerfile and no seperate line for click-odoo, click-odoo 1.3.0 gets installed instead of 1.4.2.

so for now the 1.4.2 seems to be the correct version for click-odoo (the version including the parse-version patch). it is only used in the Odoo 13.0 Dockerfile though.

pedrobaeza:
My colleague @Yajo is going to check

ap-wtioit:
@Yajo i changed the pip install command to use the released version of click-odoo-contrib (that now (1.8.0) includes the patches)

which should be equal to reverting f97341b22563f1071c8ee3fd473048187fcb8324 and c6f3bb010d27ea82307958bc0f990c7e8ce49f88

for new images currently we install click-odoo-contrib 1.8.0 and click-odoo 1.4.2 (i did not pin those versions)

PR #251, opened 2019-11-13
Add -l option to useradd

bretif:
I currently face up to an issue with current docker image.
My local user is freeipa user (not in /etc/passwd).

The fix is to add an option to useradd.
the -l flag of useradd prevents adding the user to the lastlog and faillog databases. Then no issue with overlay2 and lastlog file.

cf https://www.reddit.com/r/docker/comments/9jy0au/weird_behavior_while_creating_user_in_dockerfile/

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -192,7 +192,7 @@ ONBUILD ARG GID=1000
 
 # Enable Odoo user and filestore
 ONBUILD RUN groupadd -g $GID odoo \
-    && useradd -md /home/odoo -s /bin/false -u $UID -g $GID odoo \
+    && useradd -l -md /home/odoo -s /bin/false -u $UID -g $GID odoo \
     && mkdir -p /var/lib/odoo \
     && chown -R odoo:odoo /var/lib/odoo /qa/artifacts \
     && chmod a=rwX /qa/artifacts \
--- a/12.0.Dockerfile
+++ b/12.0.Dockerfile
@@ -188,7 +188,7 @@ ONBUILD ARG GID=1000
 
 # Enable Odoo user and filestore
 ONBUILD RUN groupadd -g $GID odoo \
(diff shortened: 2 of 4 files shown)


Yajo:
What relation has your local user with the user in the dockerfile?

bretif:
> What relation has your local user with the user in the dockerfile?

I do not have clear explanation. I use doodba for several months without any issues.
Yesterday I test it on a new PC (Ubuntu 18.04) and building my odoo image as usual fill my partition. Building with doodba take very long, failed and a temporary image of 300Gb was generated!
After investigation I see that in the "volume" in /var/lib/docker/overlay2/"VOLUMEXXX" increase because of a lastlog file.

Then I searched google and I found other people telling that when UID is very high size of container become increase exponentially.
In my case my id is very high.

```
$ id           
uid=696700501(bretif) gid=696600003(sdkadmin) groupes=696600003(sdkadmin),4(adm),24(cdrom),27(sudo),30(dip),46(plugdev),116(lpadmin),126(sambashare),131(libvirt),132(libvirt-dnsmasq),999(docker),64055(libvirt-qemu),696600000(admins),696600016(psiadmin),696600028(grafana_users),696600029(grafana_admin),696650500(sdk_techs)
```

If I build with a local user with a low id I do not have the issue.

Some resources telling to use "-l option"
- [Docker image sizes for different user ids?](https://forums.docker.com/t/docker-image-sizes-for-different-user-ids/46073/2)
- [Huge (sparse) files /var/log/lastlog and /var/log/faillog trouble docker](https://github.com/sagemathinc/cocalc/issues/2287)
- [Secure User in Docker](https://github.com/sagemathinc/cocalc/issues/2287)

I did not really understand the reason of the issue but the solution found is to user useradd with option -l to avoid this issue.

Do not hesitate if you need more details.
Thanks for your great job

Bertrand

PR #263, opened 2019-12-17
Fix dependency inclusion and exclusion detection

Yajo:
Before this patch:

- When an addon was included with `--with`, its dependencies were not detected.
- When an addon was excluded with `--without` but it was a dependency of another addon that was included, it ended up in the final list (although explicitly excluded).

TT20969 TT20963

### DIFF
--- a/bin/addons
+++ b/bin/addons
@@ -94,7 +94,8 @@ try:
         core_ok = args.core and repo == CORE
         extra_ok = args.extra and repo not in {CORE, PRIVATE}
         private_ok = args.private and repo == PRIVATE
-        if private_ok or core_ok or extra_ok:
+        manual_ok = addon in addons
+        if private_ok or core_ok or extra_ok or manual_ok:
             addon_path = os.path.join(SRC_DIR, repo, addon)
             manifest = {}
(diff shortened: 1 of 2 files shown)


pedrobaeza:
We have talked about the second point not being convenient.

Yajo:
No, that's another thing.

pedrobaeza:
I don't see in what's different. Can you explain me?

Yajo:
Yes, you were saying that if we exclude an addon, addons that depend on it shouldn't be excluded (which, in turn, would make Odoo end up installing the excluded addon because it's a dependency of another one that we're installing).

The fix here is that if an excluded addon is a dependency of a matched addon, it should not appear as selected.

The usage is for the "install" phase. For example, imagine that `addons list --extra --dependencies` includes `website`, but you don't want to include website. Then you can use `addons list --extra --dependencies --without website`. Otherwise there's no way to exclude an addon from the result if you pass `--dependencies`.

pedrobaeza:
OK, still not understanding, but you know what I don't want to happen, so feel free to continue with that in mind.

PR #272, opened 2019-12-31
Document new RELAY_PASSWORD variable

Yajo:
This breaking change will follow https://github.com/Tecnativa/doodba-scaffolding/pull/56.

@Tecnativa TT20505


Yajo:
@pedrobaeza I cannot do the variable translation we talked before. It'd have to be present in `.env` to be able to do that, but it's in another file because it's a secret.

Yajo:
No way, sorry :confused:

pedrobaeza:
Then please do:

- Announce ASAP the breaking change in the issue.
- Change our Ansible role to put both variables on each deployment.

PR #273, opened 2020-01-10
Master allow duplicate gid

ap-wtioit:
On macOS the default UID and GID are 501 and 20. The gid 20 conflicts with debians dialout group in doodba. When building on macOS one gets a message like this:
```
groupadd: GID '20' already exists
ERROR: Service 'odoo' failed to build: The command '/bin/sh -c groupadd -g $GID odoo     && useradd -l -md /home/odoo -s /bin/false -u $UID -g $GID odoo     && mkdir -p /var/lib/odoo    && chown -R odoo:odoo /var/lib/odoo /qa/artifacts     && chmod a=rwX /qa/artifacts     && sync' returned a non-zero code: 4
```

Screenshot 2020-01-09 at 12 17 53

We fixed this by allowing the added group odoo having the same gid as any existing group and added a test for building with default macOS ids.

A side effect is that if you check the group name of a directory in the container it will return the primary group name (dialout for 20) not odoo as group name. But this fix seems less complicated than deleting the primary group, adding odoo and then adding the primary group as secondary name again.

This should not affect users not having such low gid's.

Info @wt-io-it

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -424,6 +424,35 @@ class ScaffoldingCase(unittest.TestCase):
                 ),
             )
 
+    def test_uids_mac_os(self):
+        """tests if we can build an image with a custom uid and gid of odoo"""
+        uids_dir = join(SCAFFOLDINGS_DIR, "uids_mac_os")
+        for sub_env in matrix():
+            self.compose_test(
+                uids_dir,
+                sub_env,
+                # verify that odoo user has the given ids
+                ("bash", "-c", 'test "$(id -u)" == "501"'),
+                ("bash", "-c", 'test "$(id -g)" == "20"'),
+                ("bash", "-c", 'test "$(id -u -n)" == "odoo"'),
+                # all those directories need to belong to odoo (user or group odoo/dialout)
+                (
+                    "bash",
+                    "-c",
+                    'test "$(stat -c \'%U:%g\' /var/lib/odoo)" == "odoo:20"',
+                ),
+                (
+                    "bash",
+                    "-c",
+                    'test "$(stat -c \'%U:%g\' /opt/odoo/auto/addons)" == "root:20"',
+                ),
+                (
+                    "bash",
+                    "-c",
+                    'test "$(stat -c \'%U:%g\' /opt/odoo/custom/src)" == "root:20"',
+                ),
+            )
+
     def test_default_uids(self):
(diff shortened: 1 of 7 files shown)


tests/__init__.py
```diff
@@ -424,6 +424,35 @@ def test_modified_uids(self):
...
+                sub_env,
+                # verify that odoo user has the given ids
+                ("bash", "-c", 'test "$(id -u)" == "501"'),
+                ("bash", "-c", 'test "$(id -g)" == "20"'),
+                ("bash", "-c", 'test "$(id -u -n)" == "odoo"'),
+                # all those directories need to belong to odoo (user or group odoo/dialout)
+                (
+                    "bash",
+                    "-c",
+                    'test "$(stat -c \'%U:%G\' /var/lib/odoo)" == "odoo:dialout"',
```
Yajo: Please use group numeric output for the test. It avoids wondering where that `dialout` keyword comes from.

```suggestion
                    'test "$(stat -c \'%U:%g\' /var/lib/odoo)" == "odoo:20"',
```

tests/__init__.py
```diff
@@ -424,6 +424,35 @@ def test_modified_uids(self):
...
+                ),
+                (
+                    "bash",
+                    "-c",
+                    'test "$(stat -c \'%U:%G\' /opt/odoo/auto/addons)" == "root:dialout"',
```
Yajo:
```suggestion
                    'test "$(stat -c \'%U:%g\' /opt/odoo/auto/addons)" == "root:20"',
```

tests/__init__.py
```diff
@@ -424,6 +424,35 @@ def test_modified_uids(self):
...
+                ),
+                (
+                    "bash",
+                    "-c",
+                    'test "$(stat -c \'%U:%G\' /opt/odoo/custom/src)" == "root:dialout"',
```
Yajo:
```suggestion
                    'test "$(stat -c \'%U:%g\' /opt/odoo/custom/src)" == "root:20"',
```

ap-wtioit: changed in 6bd6eeb

tests/__init__.py
ap-wtioit: changed in 6bd6eeb

tests/__init__.py
ap-wtioit: changed in 6bd6eeb

PR #275, opened 2020-01-16
implement geoip installation with account id and license key

### LINKED ISSUE #253: [RFC] Add geoIP support
If you a website with customers coming from various countries, you might want to propose a different pricelist for your shop depending on the country of the visitor. 

Odoo supports this, provided that the user's requests contains geolocalization data.  That is, if the geoip parameter of session is set. NGINX, for example, provides the geoip module http://nginx.org/en/docs/http/ngx_http_geoip_module.html that will add geolocalization parameters to the vistitor's request that I assume Odoo is able to use afterwards.

As this feature would be quite interesting for any company running the ecommerce in odoo, I think that it would be interesting if Doodba could provide support for this feature.

ap-wtioit:
fix #253


ap-wtioit:
WIP TODOs:
* use curl instead of wget (wget is not available in 10.0 and 13.0 images)

entrypoint.d/45-geoip
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+if [[ -e /etc/GeoIP.conf ]] ; then
+    set -e
+    geoipupdate
```
Yajo: What if this fails to connect?

Would this slow down odoo boot process? :thinking:

build.d/750-geoip-install
```diff
@@ -0,0 +1,36 @@
...
+
+# Replace YOUR_LICENSE_KEY_HERE with an active license key associated
+# with your MaxMind account.
+LicenseKey ${GEOIP_LICENSE_KEY}
+
+# \`EditionIDs\` is from your MaxMind account.
+EditionIDs GeoLite2-City
+    " > /etc/GeoIP.conf
+    curl --silent -L --output geoipupdate_${GEOIP_UPDATER_VERSION}_linux_amd64.deb https://github.com/maxmind/geoipupdate/releases/download/v${GEOIP_UPDATER_VERSION}/geoipupdate_${GEOIP_UPDATER_VERSION}_linux_amd64.deb
+    dpkg -i --force-confold geoipupdate_${GEOIP_UPDATER_VERSION}_linux_amd64.deb
```
Yajo: I see what you're doing here and I wonder... wouldn't it be better to always supply this dependency installed instead, in the base image?

Then just let `geoipupdate` fail or not, depending on if it's properly configured or not.

build.d/750-geoip-install
```diff
@@ -0,0 +1,36 @@
...
+    rm geoipupdate_${GEOIP_UPDATER_VERSION}_linux_amd64.deb
+    chown -R root:odoo /usr/share/GeoIP/
+    chmod g+w /usr/share/GeoIP/
```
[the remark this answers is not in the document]
ap-wtioit: So we can download the updates as odoo user (in entrypoint.d and to give us the possibility to have an odoo cron running geoipupdate)

build.d/750-geoip-install
```diff
@@ -0,0 +1,36 @@
...
+
+    if [[ ! -e /opt/odoo/common/conf.d/60-geoip.conf ]] ; then
+        echo "
+[options]
+# This product includes GeoLite2 data created by MaxMind, available from https://www.maxmind.com
+geoip_database=/usr/share/GeoIP/GeoLite2-City.mmdb
```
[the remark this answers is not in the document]
ap-wtioit: The idea here was that one can overwrite the 60-geoip.conf in scaffolding and define something different if needed.

build.d/750-geoip-install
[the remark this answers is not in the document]
Yajo: Is there a possibility to put this into a folder where odoo already has permission?

Ideally: `/opt/odoo/auto/geoip/`

build.d/750-geoip-install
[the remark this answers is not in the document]
Yajo: Actually that would be the default behavior. Any time you write a config that already exists, the generated file always gets the last value.

build.d/750-geoip-install
ap-wtioit: i guess that's a possibility. Would make the base image larger but the updater is under Apache / MIT that is possible. I'll add it to the Dockerfiles then.

entrypoint.d/45-geoip
ap-wtioit: I need to test what happens in case of a connection failure. I assumed it will not let you start Odoo then. If you have no other mechanism in place this was my approach to make sure geoip is updated. For me the whole point of them changing their license from CC-BY-SA 4.0 to their custom EULA is that they can legally make sure you are updating the database frequently.
My intention was to make it as easy as it can be to comply with the EULA.

And yes running code in entrypoint makes the boot process slower (but i guess geoipupdate is faster than any odoo stuff ;-) )

build.d/750-geoip-install
Yajo: Yes, please. Whenever possible, it's a decision in this project to inflate the base image, as it makes subimages build faster.

entrypoint.d/45-geoip
Yajo: For example, demos are supplied with network isolation out of the box as a doodba feature. Indeed we need to cover this possibility.

First of all, just check:

- what happens if there's no network? like in demos
- how long it takes when there is network?

Depending on the results, we'll know if there's a real problem.

build.d/750-geoip-install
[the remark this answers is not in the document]
ap-wtioit: As far as i understood it that only works on variable level. Like a=b in doodba and a=c in scaffolding would set a=c. But if you want to get rid of the variable definition like no a=... statement in the config this is currently not possible to do that in scaffolding if its already defined in doodba. So the idea was that if you do not want deoip_database defined with any value you still could do that by providing an empty (just with `[options]`) 60-geoip.conf in your scaffolding.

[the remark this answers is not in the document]
Yajo: Well, I think we're not understanding each other.

See, my question was about if odoo supports setting an unexisting file in this setting. It seems to be supported since v8+ (where geoip support started) and [all it would do is to produce a warning](https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/addons/website/models/ir_http.py#L96). Other versions behave similarly, and this seems a good behavior, so please just add the raw `conf.d/60-geoip.conf` file (you don't need to create it or not depending on circumstances).

ap-wtioit:
> I'm confused. Are you using the settings in the build or in the entrypoint phase?

I'm enabling enabling the geoip databaase during both phases. So it would be possible to build the credentials into the image or to enable it via environment parameters.

build.d/750-geoip-install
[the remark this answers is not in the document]
ap-wtioit: I added the raw file, also it seems there is always a default set in the config https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/openerp/tools/config.py#L306, so you're right it doesn't matter that we always add 60-geoip.conf.

build.d/750-geoip-install
ap-wtioit: I added the geoip updater to the base image.

entrypoint.d/45-geoip
ap-wtioit: I still need to test that. What do you mean by "like in demos" (i cannot find the reference demo in https://github.com/Tecnativa/doodba/blob/master/README.md nor in https://github.com/Tecnativa/doodba-scaffolding)?

Yajo: I mean https://github.com/Tecnativa/doodba#testing, where it says "it is isolated".

Yajo:
IMHO that's not necessary. It's better to just have 1 way of doing things. AFAICS, this could be better done in the build phase, where the job runs as root and thus we can leak less secrets to the odoo user. Also, if the geoip database is filled at build time, you don't need to slow down the entrypoint or check for network problems, as the build phase never has network isolation.

ap-wtioit:
> ... Also, if the geoip database is filled at build time, you don't need to slow down the entrypoint or check for network problems, as the build phase never has network isolation.

Well we need a way to update geoip while running the containers to enable the users of geoip to comply with the GeoIP EULA that requires you to keep an eye on the list of ips that need to be removed from the database. And the easiest way is to give the odoo user the rights to run the geoipupdate command. Thats why there is a test that the geoipupdate command works in a running container.

For me the possibility to define the credentials during build time would be benefitial for the CI instances that i'm running for a short period of time that are in our control and where i can be responsible for complying with the EULA. On the other hand when we would deliver images to our customers or their server the customer should be responsible for complying with the EULA and should be able to enter their credentials at runtime. Thats why i implemented both ways.

Yajo:
I see, hat makes sense :thinking:. However if that's the case, then we can implement this based only on env variables, as you can use those also in your CI to test this behavior.

Let's remove stuff in the `build.d` step and focus on just base image and entrypoint. There's still https://github.com/Tecnativa/doodba/pull/275#discussion_r367399814 pending though...

entrypoint.d/45-geoip
ap-wtioit: Timing tests:
* ran with `time geoipupdate`
* at our office speedtest.net: 6ms ping, 300Mbps down, 50Mbps up
* ping updates.maxmind.com average 11.739ms

Result:
* with network connectivity 0,542s
* with no network connectivity (docker compose network internal true) 20,005s

Error with timeout, so it seems there is a 20 second timeout:
`error retrieving updates: error retrieving filename: error performing HTTP request: Get https://updates.maxmind.com/app/update_getfilename?product_id=GeoLite2-Country: dial tcp: lookup updates.maxmind.com on 127.0.0.11:53: read udp 127.0.0.1:37138->127.0.0.11:53: i/o timeout`

ap-wtioit:
still need to test v9+ and v8 (and add dependencies)

entrypoint.d/45-geoip
Yajo: Yikes 20s is a lot... :confused: 

Try to spawn the process, it should not affect boot time:

```suggestion
    geoipupdate &
```

Please test and report :blush:

conf.d/60-geoip.conf
```diff
@@ -0,0 +1,3 @@
+[options]
+# This product includes GeoLite2 data created by MaxMind, available from https://www.maxmind.com
```
Yajo: Actually we don't include data, so it's better not to lie IMHO.

```suggestion
```

entrypoint.d/45-geoip
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+if [[ "$GEOIP_ACCOUNT_ID" != "" && "$GEOIP_LICENSE_KEY" != "" ]] ; then
```
Yajo: I'm not sure but I think this won't do what you expect; try:

```suggestion
if [ -n "$GEOIP_ACCOUNT_ID" -a -n "$GEOIP_LICENSE_KEY" ] ; then
```

Docs: http://www.tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html

entrypoint.d/45-geoip
```diff
@@ -0,0 +1,15 @@
...
+    set -e
+    # see https://dev.maxmind.com/geoip/geoipupdate/
+    log INFO "Activating GeoIP/GeoLite2 updates"
+    log INFO "This product includes GeoLite2 data created by MaxMind, available from https://www.maxmind.com"
+    sed 's/^AccountID\s.*$/AccountID '"$GEOIP_ACCOUNT_ID"'/g;s/^LicenseKey\s.*$/LicenseKey '"$GEOIP_LICENSE_KEY"'/g' /etc/GeoIP.conf > /tmp/GeoIP.conf
```
Yajo: Add -i and sed replaces in place AFAIK.

```suggestion
    sed -i 's/^AccountID\s.*$/AccountID '"$GEOIP_ACCOUNT_ID"'/g;s/^LicenseKey\s.*$/LicenseKey '"$GEOIP_LICENSE_KEY"'/g' /etc/GeoIP.conf
```

Then you can remove the next 2 lines also.

tests/__init__.py
```diff
@@ -20,6 +20,9 @@
 )
 PG_VERSIONS = frozenset(environ.get("PG_VERSIONS", "12").split())
 SCAFFOLDINGS_DIR = join(DIR, "scaffoldings")
+GEIOP_CREDENTIALS_PROVIDED = environ.get("GEOIP_LICENSE_KEY", False) and environ.get(
```
Yajo:
```suggestion
GEIOP_CREDENTIALS_PROVIDED = environ.get("GEOIP_LICENSE_KEY") and environ.get(
```

tests/__init__.py
```diff
@@ -20,6 +20,9 @@
...
+    "GEOIP_ACCOUNT_ID", False
```
Yajo:
```suggestion
    "GEOIP_ACCOUNT_ID"
```

entrypoint.d/45-geoip
```diff
@@ -0,0 +1,15 @@
...
+    cat /tmp/GeoIP.conf > /etc/GeoIP.conf
+    rm /tmp/GeoIP.conf
+fi
+
+if [[ "$(grep 'AccountID' /etc/GeoIP.conf | sed 's/AccountID //')" != "0" ]] ; then
```
Yajo: Sorry about this, I'm just not familiar with the other syntax and I prefer to keep it consistent with the rest of the codebase.

```suggestion
if [ "$(grep 'AccountID' /etc/GeoIP.conf | sed 's/AccountID //')" != "0" ] ; then
```

Yikes, I should have moved to plumbum long ago here...

entrypoint.d/45-geoip
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+if [[ "$GEOIP_ACCOUNT_ID" != "" && "$GEOIP_LICENSE_KEY" != "" ]] ; then
+    set -e
```
Yajo: I see you repeat it twice... IMHO it's better to add it just once at the top of the file.

tests/scaffoldings/geoip/custom/docker-compose.yaml
```diff
@@ -0,0 +1,35 @@
+version: "2.1"
```
Yajo: AFAICS this file is not being used, right? If so, remove it please.

entrypoint.d/45-geoip
ap-wtioit: well, i had `-i` here before, with `-i` sed tries to create a temp file in the directory the file is in. With /etc/ this does not work. I found no convenient way to convince sed to place the temporary file in a writeable directory. So this was the suggested workarround for the problem.

tests/__init__.py
ap-wtioit: I prefer to have explicit return values for dict get to avoid type confusion bugs with implicit return values. Even if it wouldn't be needed here.

tests/__init__.py
ap-wtioit: I prefer to have explicit return values for dict get to avoid type confusion bugs with implicit return values. Even if it wouldn't be needed here.

entrypoint.d/45-geoip
ap-wtioit: i use one `set -e` now

entrypoint.d/45-geoip
Yajo: Oh ok then, thanks for the explanation.

conf.d/60-geoip.conf
ap-wtioit: Fixed it. Its also better not to lie in my opinion. Thanks for noticing the error.

ap-wtioit:
Tested V8.0 after installing dependencies:
* apt: build-essential libgeoip-dev python-dev 
* pip: GeoIP
we get the following error:
`Invalid database type GeoIP Country Edition, expected GeoIP City Edition, Rev 1`
and no data

so geoip doesn't work in odoo v8 anymore (only if you have a backed up v1 geoip database somewhere)

Yajo:
This is still red and has some pending comments (kind reminder).

entrypoint.d/45-geoip
ap-wtioit: i tested the `&` (running the geoipupdate in background).


test commands:
with disabled maxmind
```
```

Yajo: Can you confirm also that quickly stopping the odoo container reaps the geoipupdate process properly instead of leaving it as zombie?

ap-wtioit: yes it seems it is killed as soon as the main process (odoo) stops

tested with `docker-compose up` in one window and `while true; do date; ps fax | grep -E '(geoipupdate|odoo)' | grep -v grep; sleep 1; done` in the other pressing CTRL+C in the docker window as soon as odoo process is started.

it takes about 5 seconds for odoo to stop and geoiupdate is gone as soon as odoo is gone in ps fax

Yajo: Perfect! Thanks!

tests/__init__.py
```diff
@@ -277,6 +280,10 @@ def test_smallest(self):
             ("ssh", "-V"),
             # We are able to dump
             ("pg_dump", "-f/var/lib/odoo/prod.sql", "prod"),
+            # Geoip should not be activated
+            ("bash", "-xc", 'test "$(which geoipupdate)" != ""'),
```
Yajo:
```suggestion
            ("which", "geoipupdate"),
```

entrypoint.d/45-geoip
Yajo: I'm having a hard time understanding this `if`. Could you explain what you are trying to do please?

I'm tired of bash :man_facepalming: , when you explain it to me I'm going to upload a commit translating this into plumbum and you review it please. Thanks!

build.d/750-geoip-install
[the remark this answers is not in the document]
ap-wtioit: i will change that and add symlinks into the image to allow calling `geoipupdate` without args as default. without symlinks we (and the users) would need to call `geoipupdate -f  /opt/odoo/auto/geoip/geoip.conf -d /opt/odoo/auto/geoip` for it to work. so i'll try to put symlinks for the default geoip files / directories pointing to /opt/odoo/auto/geoip and report back

8.0.Dockerfile
```diff
@@ -82,6 +86,7 @@ RUN pip install \
         virtualenv \
         wdb \
         simplejson \
+        geoip2 \
```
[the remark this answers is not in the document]
ap-wtioit: nope, i did say it is not working for v8 (because GeoIP is no longer working with current data, that is only available in geoip2 format)

the v8 dockerfile is used for v7, v8, v9, v10 so to get it working in v9 and v10 i need to add the dependency into the v8 file:
```
lrwxrwxrwx  1 bigbear3001 bigbear3001    14 Mai 14  2018 10.0.Dockerfile -> 8.0.Dockerfile
-rw-rw-r--  1 bigbear3001 bigbear3001  8109 Feb 19 13:23 11.0.Dockerfile
-rw-rw-r--  1 bigbear3001 bigbear3001  7775 Feb 19 14:49 12.0.Dockerfile
-rw-rw-r--  1 bigbear3001 bigbear3001  7643 Feb 19 15:18 13.0.Dockerfile
lrwxrwxrwx  1 bigbear3001 bigbear3001    14 Jän 13 10:56 7.0.Dockerfile -> 8.0.Dockerfile
-rw-rw-r--  1 bigbear3001 bigbear3001  8398 Feb 19 13:25 8.0.Dockerfile
lrwxrwxrwx  1 bigbear3001 bigbear3001    14 Mai 14  2018 9.0.Dockerfile -> 8.0.Dockerfile
```

README.md
```diff
@@ -653,6 +654,65 @@ about it. Just execute `odoo` and it will work fine.
...
+    odoo:
+        ...
+        depends_on:
+            ...
+            - maxmind_proxy
+            ...
+        ...
+    ...
+
+    maxmind_proxy:
```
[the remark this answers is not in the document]
ap-wtioit: Well my idea was still to keep it optional (for the purist that only want GPL/Apache2 stuff in their images) and therefore we wouldn't need the proxy for all devel and test environments.

Also the test environment currently has no whitelist proxies defained as a default but only the comment in the readme that one should create the globalwhitelist_shared network with proxies in it:
https://github.com/Tecnativa/doodba-scaffolding/blob/master/test.yaml
BTW we propably should have a detailed explanation how the whitelist proxy network is supposed to work (as i couldn't figure it out when i started working with doodba)

tests/__init__.py
ap-wtioit: The difference between the two lines is that bash -xc also prints the invoked command making it easier to debug if it fails.

entrypoint.d/45-geoip
ap-wtioit: Searching the line in /etc/GeoIP.conf (to be moved) that contains "AcountID " getting rid of the "AccountID " part, and checking if the set AccountID is not 0 (the default value)
`AccountID 1234` is the line when it is set, `AccountID 0` is the line when it is not set

README.md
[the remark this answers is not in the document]
Yajo: OK, you're true.

Then add it here please: https://github.com/Tecnativa/doodba/#global-whitelist

entrypoint.d/45-geoip
```diff
@@ -0,0 +1,29 @@
...
+    )
+    # Obtain current GeoIP configuration
+    geoip_file = local.path("/etc/GeoIP.conf")
+    geoip_conf = geoip_file.read()
+    # Patch it with provided account and license keys
+    geoip_conf = re.sub(r"^AccountID\s*.*$", "AccountID {}".format(GEOIP_ACCOUNT_ID))
+    geoip_conf = re.sub(r"^LicenseKey\s*.*$", "LicenseKey {}".format(GEOIP_LICENSE_KEY))
+    geoip_file.write(geoip_conf)
+    # run geoip update in background to make sure it doesn't block odoo startup
+    geoipupdate & NOHUP(stdout=sys.stdout, stderr=sys.stderr)
```
Yajo: Sorry to bother you again my friend.

Could you please make sure again this works as expected and detaches the geoipupdate process, lets it continue even when the entrypoint script is gone, and is not left as zombie if odoo stops before geoip ends updating?

I just want to make sure this works as expected :sweat_smile: 

BTW you see here the same script in python + plumbum. :sparkles: :heart_eyes: :sparkles:

README.md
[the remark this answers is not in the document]
ap-wtioit: i added it to the global-whitelist template in the readme

entrypoint.d/45-geoip
ap-wtioit: I fixed the missing argument for re.sub in https://github.com/Tecnativa/doodba/pull/275/commits/8518587aa41a6c7f9d3738800b3395bd24d630b0 but i still get errors:
```
doodba INFO: Waiting until postgres is listening at db...
doodba INFO: Unaccent extension installed
doodba INFO: Linking all addons from /opt/odoo/custom/src/addons.yaml in /opt/odoo/auto/addons
doodba INFO: Activating GeoIP/GeoLite2 updates
doodba INFO: This product includes GeoLite2 data created by MaxMind, available from https://www.maxmind.com
  File "/opt/odoo/common/entrypoint.d/45-geoip", line 29, in <module>
    geoipupdate & NOHUP(stdout=sys.stdout, stderr=sys.stderr)
  File "/usr/local/lib/python3.7/site-packages/plumbum/commands/modifiers.py", line 377, in __rand__
    return cmd.nohup(cmd, self.cwd, stdout, self.stderr, append)
  File "/usr/local/lib/python3.7/site-packages/plumbum/commands/base.py", line 152, in nohup
    return self.machine.daemonic_popen(self, cwd, stdout, stderr, append)
  File "/usr/local/lib/python3.7/site-packages/plumbum/machines/local.py", line 337, in daemonic_popen
    return posix_daemonize(command, cwd, stdout, stderr, append)
  File "/usr/local/lib/python3.7/site-packages/plumbum/commands/daemons.py", line 70, in posix_daemonize
    raise ProcessExecutionError(argv, rc, "", output)
plumbum.commands.processes.ProcessExecutionError: Unexpected exit code: 256
Command line: | /usr/bin/geoipupdate
Stderr:       | Traceback (most recent call last):
              |   File "/usr/local/lib/python3.7/site-packages/plumbum/commands/daemons.py", line 39, in posix_daemonize
              |     stdout = open(stdout, "a" if append else "w")
              | TypeError: expected str, bytes or os.PathLike object, not _io.TextIOWrapper
  File "/opt/odoo/common/entrypoint", line 32, in <module>
  File "/usr/local/lib/python3.7/subprocess.py", line 347, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '/opt/odoo/common/entrypoint.d/45-geoip' returned non-zero exit status 1.
```

does this mean exit code 256 from geoipupdate or does this mean something else?

ap-wtioit: also fixed the regexes now (still getting the same error)

ap-wtioit: ok i give up, i cannot find NOHUP in the plumbum documentation
the problem seems to be the `geoipupdate & NOHUP(stdout=sys.stdout, stderr=sys.stderr)`
i tested on a python console with the following:
```
bigbear3001@wt-io-it-bigbear3001:~$ python3 -c 'from plumbum import NOHUP, BG; from plumbum.cmd import bash; import sys; bash["-c", "sleep 1; echo 'finished'"] & NOHUP(stdout=sys.stdout)'
  File "<string>", line 1, in <module>
  File "/home/bigbear3001/.local/lib/python3.6/site-packages/plumbum/commands/modifiers.py", line 377, in __rand__
    return cmd.nohup(cmd, self.cwd, stdout, self.stderr, append)
  File "/home/bigbear3001/.local/lib/python3.6/site-packages/plumbum/commands/base.py", line 152, in nohup
    return self.machine.daemonic_popen(self, cwd, stdout, stderr, append)
  File "/home/bigbear3001/.local/lib/python3.6/site-packages/plumbum/machines/local.py", line 337, in daemonic_popen
    return posix_daemonize(command, cwd, stdout, stderr, append)
  File "/home/bigbear3001/.local/lib/python3.6/site-packages/plumbum/commands/daemons.py", line 70, in posix_daemonize
    raise ProcessExecutionError(argv, rc, "", output)
plumbum.commands.processes.ProcessExecutionError: Unexpected exit code: 256
Command line: | /bin/bash -c 'sleep 1; echo finished'
Stderr:       | Traceback (most recent call last):
              |   File "/home/bigbear3001/.local/lib/python3.6/site-packages/plumbum/commands/daemons.py", line 39, in posix_daemonize
              |     stdout = open(stdout, "a" if append else "w")
              | TypeError: expected str, bytes or os.PathLike object, not _io.TextIOWrapper
```
but BG seems to do the trick:
```
bigbear3001@wt-io-it-bigbear3001:~$ python3 -c 'from plumbum import NOHUP, BG; from plumbum.cmd import bash; import sys; bash["-c", "sleep 1; echo 'finished'"] & BG(stdout=sys.stdout)'
bigbear3001@wt-io-it-bigbear3001:~$ finished
```

ap-wtioit: Retested after the latest commit: https://github.com/Tecnativa/doodba/pull/275/commits/495c6bfc5924dfcd58d43e3daca6b9e6e53ec9f9

* geoipupdate does not block odoo startup
* geoipupdate terminates as soon as odoo terminates when running `docker-compose up` and pressing CTRL+C
* geoipupdate error message is shown in `docker-compose logs` of odoo container (it does not fail silently)
* after starting maxmind_proxy odoo needs a restart / workers need to restart to activate geoip functionality

ap-wtioit:
TODOs:
* move geoip data to /opt/odoo/auto/geoip
* enable tests on TravisCI (needs GEOIP_ACCOUNT_ID and GEOIP_LICENSE_KEY set in travis variables by admin)
* move geoip install into first install block (on all images)

entrypoint.d/45-geoip
Yajo: Excellent!

NOHUP is documented here: https://plumbum.readthedocs.io/en/latest/local_commands.html#background-and-foreground

According to the docs, I thought that using BG would stop the background job as soon as the main job exited, but it seems to not do that, so... perfect!

Thanks a lot.

Yajo:
> enable tests on TravisCI (needs GEOIP_ACCOUNT_ID and GEOIP_LICENSE_KEY set in travis variables by admin)

I'll handle this for the master branch. I cannot enable it in PRs because of security.

13.0.Dockerfile
```diff
@@ -87,6 +91,12 @@ RUN python -m venv --system-site-packages /qa/venv \
     && mkdir -p /qa/artifacts \
     && git clone --depth 1 $MQT /qa/mqt
 
+# Install GeoIP updater
+ARG GEOIP_UPDATER_VERSION=4.1.5
+RUN curl --silent -L --output geoipupdate_${GEOIP_UPDATER_VERSION}_linux_amd64.deb https://github.com/maxmind/geoipupdate/releases/download/v${GEOIP_UPDATER_VERSION}/geoipupdate_${GEOIP_UPDATER_VERSION}_linux_amd64.deb \
+    && dpkg -i geoipupdate_${GEOIP_UPDATER_VERSION}_linux_amd64.deb \
+    && rm geoipupdate_${GEOIP_UPDATER_VERSION}_linux_amd64.deb
```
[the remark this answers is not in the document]
ap-wtioit: I moved the geoipupdate install to the first install were wkthmltopdf also gets installed

ap-wtioit:
@Yajo i hope you are well. I think i fixed all the issues. If it is ok with you i would rebase this one to prepare for a merge.

11.0.Dockerfile
```diff
@@ -29,6 +29,11 @@ ENV DB_FILTER=.* \
     WDB_WEB_PORT=1984 \
     WDB_WEB_SERVER=localhost
 
+# Enable configuring geoip with scaffolding environment (by setting GEOIP_ACCOUNT_ID and GEOIP_LICENSE_KEY)
+ENV GEOIP_ACCOUNT_ID="" \
+    GEOIP_LICENSE_KEY=""
+ARG GEOIP_UPDATER_VERSION=4.1.5
```
Yajo: Please move ARGs to the top (see where the others are).

Then please join that new ENV with the ENV clause above.

Try to sort them alphabetically.

This is to make good usage of cache and reduce layers.

12.0.Dockerfile
```diff
@@ -29,6 +29,11 @@ ENV DB_FILTER=.* \
     WDB_WEB_PORT=1984 \
     WDB_WEB_SERVER=localhost
 
+# Enable configuring geoip with scaffolding environment (by setting GEOIP_ACCOUNT_ID and GEOIP_LICENSE_KEY)
+ENV GEOIP_ACCOUNT_ID="" \
+    GEOIP_LICENSE_KEY=""
+ARG GEOIP_UPDATER_VERSION=4.1.5
```
Yajo: Same as above.

13.0.Dockerfile
```diff
@@ -29,6 +29,11 @@ ENV DB_FILTER=.* \
     WDB_WEB_PORT=1984 \
     WDB_WEB_SERVER=localhost
 
+# Enable configuring geoip with scaffolding environment (by setting GEOIP_ACCOUNT_ID and GEOIP_LICENSE_KEY)
+ENV GEOIP_ACCOUNT_ID="" \
+    GEOIP_LICENSE_KEY=""
+ARG GEOIP_UPDATER_VERSION=4.1.5
```
Yajo: Same as above.

8.0.Dockerfile
```diff
@@ -29,6 +29,11 @@ ENV DB_FILTER=.* \
     WDB_WEB_PORT=1984 \
     WDB_WEB_SERVER=localhost
 
+# Enable configuring geoip with scaffolding environment (by setting GEOIP_ACCOUNT_ID and GEOIP_LICENSE_KEY)
+ENV GEOIP_ACCOUNT_ID="" \
+    GEOIP_LICENSE_KEY=""
+ARG GEOIP_UPDATER_VERSION=4.1.5
+
```
Yajo: Same as above.

entrypoint.d/45-geoip
```diff
@@ -0,0 +1,37 @@
...
+    geoip_conf = re.sub(
+        r"(^|\n)AccountID\s*[^\n]*(\n|$)",
+        "\\1AccountID {}\\2".format(GEOIP_ACCOUNT_ID),
+        geoip_conf,
+    )
+    geoip_conf = re.sub(
+        r"(^|\n)LicenseKey\s*[^\n]*(\n|$)",
+        "\\1LicenseKey {}\\2".format(GEOIP_LICENSE_KEY),
+        geoip_conf,
+    )
```
Yajo: Before we had this:

```suggestion
    geoip_conf = re.sub(r"^AccountID\s*.*$", "AccountID {}".format(GEOIP_ACCOUNT_ID))
    geoip_conf = re.sub(r"^LicenseKey\s*.*$", "LicenseKey {}".format(GEOIP_LICENSE_KEY))
```

It seems more readable. Wasn't it enough? Keep in mind that AFAIK `^` and `$` match newlines if the regexp is not in multiline mode.

Yajo:
> Don't mind about rebasing, I'll squash on merge for your comfort if you want.

Oops I missed the point that there are conflicts. Then please rebase.

entrypoint.d/45-geoip
ap-wtioit: as far as i remember this did not work:
* re.sub needs a third argument (the source string)
* your regex didn't match LicenseKey and AccoundID in all places it could appear (regex ^ applied on string only matches the beginning of the string (not beginning of a line), the same for `$`(end of string not newline) )

8.0.Dockerfile
ap-wtioit: i moved it to the top

13.0.Dockerfile
ap-wtioit: i moved it to the top

12.0.Dockerfile
ap-wtioit: i moved it to the top

11.0.Dockerfile
ap-wtioit: i moved it to the top

PR #276, opened 2020-01-20
Added config-expand script in entrypoint to allow more control over o…

PCatinean:
Whenever there's a parameter to change in the odoo config as of now if I'm not mistaken we have to add them one by one from ENV to the config. This is what I use in my deployments to k8s clusters to make it easier and flexible at the same time. Let me know if this is of any use in the general project.


Yajo:
Hi there, thanks for the PR.

It seems a pretty intelligent way to handle an old problem.

What's missing:

1. Add this logic into https://github.com/Tecnativa/doodba/blob/master/bin/config-generate directly.
1. Provide a deprecation path to the old system where it's duplicated with the new one.

PCatinean:
Yeah I just uploaded mine from scaffolding first to see if it's of any use. I'll do my best to get to it this week

gdgellatly:
Just a question, does this cause issues with non odoo options that would usually sit in different parts of the conf file? e.g. the old queue job settings.

joao-p-marques:
> Just a question, does this cause issues with non odoo options that would usually sit in different parts of the conf file? e.g. the old queue job settings.

AFAICS it shouldn't cause any problem... All this does is append to the parser (or replace) the variables defined by environment, so you should still be able to add a custom config section to the file and use the variables for general Odoo settings at the same time.

joao-p-marques (changes requested):
However, maybe this deserves a test, possibly a small change to https://github.com/Tecnativa/doodba/tree/master/tests/scaffoldings/settings
@PCatinean can you do it, please? 

Thanks for the nice improvement :smiley:

bin/config-generate
```diff
@@ -3,23 +3,31 @@
...
+    sys.path.append(ODOO_DIR)
+    from odoo.tools.config import configmanager
+
 try:
     from configparser import RawConfigParser
 
     parser = RawConfigParser(strict=False)
 except ImportError:
-    # Python 2, where strict=True doesn't exist
+    # Python 2, where strict=True doesn"t exist
```
joao-p-marques, line 30: Also, could you change this back to `doesn't`?

PR #277, opened 2020-01-20
add missing step for inverseproxy

yelizariev:
Without the step you will get following error:

ERROR: Network inverseproxy_shared declared as external, but could not be found. Please create the network manually using `docker network create inverseproxy_shared` and try again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -743,12 +743,6 @@ safe:
 - `./.docker/smtp.env` must define `RELAY_PASSWORD` (password to access the real SMTP relay).
 - `./.docker/backup.env` must define `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` (obtained from S3 provider) and `PASSPHRASE` (to encrypt backup archives).
 
-###### Booting production
-
-Once you fixed everything needed, run it with:
-
-    docker-compose -f prod.yaml up --build --remove-orphans
-
 ###### Global inverse proxy
 
 For [production][] and [testing][] templates to work fine, you need to have a
@@ -843,6 +837,12 @@ This allows you to:
 - Have multiple Odoo instances in each node.
 - Add an SSL layer automatically and for free.
 
+###### Booting production
+
+Once you fixed everything needed and started [global inverse proxy](#global-inverse-proxy), run the production environment with:
+
+    docker-compose -f prod.yaml up --build --remove-orphans
+
 ##### Testing
 
 A good rule of thumb is test in testing before uploading to production, so this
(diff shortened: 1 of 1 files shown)


yelizariev:
I miss somtihng. I'm getting following error on trying to start inverproxy

    ERROR: Network "inverseproxy_shared" needs to be recreated - option "encrypted" has changed

adding ``--opt encrypted`` doesn't help.

@Yajo could you advice?

yelizariev:
ok, this seems work:


    --opt encrypted=1 --internal

README.md
```diff
@@ -747,6 +747,7 @@ safe:
 
 Once you fixed everything needed, run it with:
 
+    docker network create inverseproxy_shared --opt encrypted=1 --internal
     docker-compose -f prod.yaml up --build --remove-orphans
 
 ###### Global inverse proxy
```
Yajo: If you follow instructions in this section, the `inverseproxy_shared` network is created for you.

Do we need anything else?

HviorForgeFlow:
You can use --remove-orphans anytime you stop one stage and go to another.

Yajo:
That helps for containers, not networks AFAIK.

HviorForgeFlow:
Yes, you just need to ensure you stopped container in previous stage using down.

README.md
yelizariev: As I understand, external networks are not created automatically. So, explicit creation is required.

I'm getting following error after executing `    docker-compose -f prod.yaml up --build --remove-orphans`:

    ERROR: Network inverseproxy_shared declared as external, but could not be found. Please create the network manually using `docker network create inverseproxy_shared` and try again.

yelizariev: Sorry, I probably misunderstood you. If you are asking are there more things to do to make it run, then I can say that I don't know about it. These two steps seems working to run production with our without inverse proxy (though I tested it with traefik 2.1)

Yajo: Yes, there seems to be a misunderstanding here.

I mean, do this:

1. Follow instructions in section [Global inverse proxy][].
1. At this point, the `inverseproxy_shared` network has been automatically created for you, with a working traefik deployment.
1. Follow instructions in section [Booting production][]

If you do this in this order, you shouldn't need to create the network manually.

So, that's why I ask if is there anything missing in the docs. Some ideas:

1. Add a FAQ about the error of the missing network, which directs users to follow instructions from [Global inverse proxy][] or either run the command you're suggesting here.
1. Put the section *[Booting production][]* below *[Global inverse proxy][]* instead of the contrary (as it is right now).

[Global inverse proxy]: https://github.com/Tecnativa/doodba/blob/5a5d94dc42e6afa7d1b29eb2ab16c11ed83924ee/README.md#global-inverse-proxy
[Booting production]: https://github.com/Tecnativa/doodba/blob/5a5d94dc42e6afa7d1b29eb2ab16c11ed83924ee/README.md#booting-production

yelizariev: Oh, I got it

> If you follow instructions in this section,

by *this* section you meant ``###### Global inverse proxy`` which is the last line of the quoted piece of code.

----
Indeed, after executing ``docker-compose -p inverseproxy -f inverseproxy.yaml up -d`` the network is created... 

Moving Booting production below Global inverse proxy with adding extra notes looks  as a good idea

yelizariev: @Yajo take a look on updated version, please

PR #282, opened 2020-01-24
fix db_filter typo in conf.d/20-database.conf

chesty:
apologies, I haven't tested this fully. 

setting DB_FILTER="^%d$" in .env didn't work for me.

I manually edited auto/odoo.conf in the container with nano and changed the line with dbfilter = to db_filter = and restart the container and it worked. I think this patch should fix it but it's not tested.


chesty:
this is wrong. it should be dbfilter -> https://www.odoo.com/documentation/13.0/setup/deploy.html

sorry about that. I've never used multi-tenant before, it works OK in my dev environment but I can't get it to work in doodba.

PR #286, opened 2020-02-04
add toc

### LINKED ISSUE #280: Readme's TOC
@Yajo @pedrobaeza 
What do you think about adding TOC?

Example: https://github.com/yelizariev/doodba/blob/toc/README.md (*doctoc* doesn't generated all links properly, I hope it will be fixed soon)

I also made a Github Actions to regenerate TOC: https://github.com/yelizariev/doodba/blob/toc/.github/workflows/doctoc.yml

yelizariev:
closes #280


yelizariev:
@Yajo I need your advice on how to debug pre-commit errors 
https://travis-ci.org/Tecnativa/doodba/jobs/645836860?utm_medium=notification&utm_source=github_status

yelizariev:
I sent an issue https://github.com/trussworks/pre-commit-hooks/issues/1

yelizariev:
Fixed by switching to another precommit hook

yelizariev:
Another precommit listed here: https://pre-commit.com/hooks
doesn't work either https://github.com/Lucas-C/pre-commit-hooks-nodejs/issues/3

yelizariev:
Ready to review
https://github.com/Tecnativa/doodba/blob/586f309fb3b709d127f2cddc7c05d68fc12de108/README.md#how

README.md
```diff
@@ -434,7 +515,7 @@ The great [`click-odoo`][] scripting framework and the collection of scripts fou
 
 \* Note: This replaces the deprecated `python-odoo-shell` binary.
 
-### [`nano`][]
+### [`nano`](https://www.nano-editor.org/)
```
Yajo: Shouldn't links from headers be removed to avoid problems?

pedrobaeza:
In OCA we put
```
.. contents::
   :local:
```
and it's auto-populated by the RST reader. Can't this be the same?

Yajo:
Nope, this is markdown.

pedrobaeza:
And why keeping it as MD instead of RST? I prefer to standardize all in the same format if possible.

Yajo:
MD gives more pleasure to write, and it's rendered fine out of the box here in Github. Also has better supoprt like everywhere (example: prettier). RST is better suited for large codebases where docs are autobuilt by code introspection, but here there's almost no code (and will never be, probably), so I'm not sure it's a good choice.

In any case, this is the MVC towards having a good TOC in our README, so it's a good enhancement. We can talk on doc refactors later.

pedrobaeza:
About the write pleasure I differ (just because I use to write RST), but the question was raised because the effort for having the TOC in MD seems to be a lot.

Yajo:
Where do you see such a big effort here? :thinking: It's just adding a pre-commit hook that autogenerates it...

pedrobaeza:
Well, @yelizariev has invested here a lot of time

Yajo:
And that's something I appreciate a lot indeed! :raised_hands: 

Now that the fix is here, we should accept it and move on. :+1: 

There's just one open question still: https://github.com/Tecnativa/doodba/pull/286#discussion_r376289039

README.md
yelizariev: It works in this case, i.e. when the link is entire header

PR #297, opened 2020-03-23
Add python lib firebase_admin for standard odoo module 

PCatinean:
social_push_notification is a standard odoo module that has this external dependency.

Not sure what the policy is here since odoo does not provide it in the requirements.txt themselves. I thought this might be helpful as it supports a standard module.

### DIFF
--- a/13.0.Dockerfile
+++ b/13.0.Dockerfile
@@ -127,6 +127,7 @@ RUN build_deps=" \
         astor \
         git-aggregator \
         click-odoo-contrib \
+        firebase_admin \
         pg_activity \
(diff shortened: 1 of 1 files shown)


pedrobaeza:
For me this should be included in specific repository installing that module.

Yajo:
Where's that module? I can't find it.

PCatinean:
Enterprise standard module (checked on runbot)

Yajo:
So it's not standard, it's enterprise. You'll have to add that dependency to your scaffolding's `pip.txt` as with any other external repo.

PR #304, opened 2020-05-14
Make msgmerge CLI tool available

Yajo:
This is a requirement to be able to use click-odoo-makepot

@Tecnativa TT23467

### DIFF
--- a/tests/scaffoldings/addons_env/custom/src/addons.yaml
+++ b/tests/scaffoldings/addons_env/custom/src/addons.yaml
@@ -8,7 +8,7 @@ odoo/addons:
   - sale
 ---
 ENV:
-  DEFAULT_REPO_PATTERN: https://github.com/it-projects-llc/{}.git
-  DEPTH_DEFAULT: 2
+  DEFAULT_REPO_PATTERN: https://github.com/itpp-labs/{}.git
+  ODOO_VERSION: 554222498d5101ef59fc4d3e1059e2b407dffd6f
 misc-addons:
   - web_debranding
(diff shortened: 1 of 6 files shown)


tests/scaffoldings/addons_env/custom/src/addons.yaml
```diff
@@ -8,7 +8,7 @@ odoo/addons:
   - sale
 ---
 ENV:
-  DEFAULT_REPO_PATTERN: https://github.com/it-projects-llc/{}.git
-  DEPTH_DEFAULT: 2
+  DEFAULT_REPO_PATTERN: https://github.com/itpp-labs/{}.git
+  ODOO_VERSION: 554222498d5101ef59fc4d3e1059e2b407dffd6f
```
pedrobaeza, line 12: Why do you pin an Odoo version? This may fail real builds while CI is green.

Yajo: The build was broken because the module we were downloading for test disappeared from 13.0 branch; and after all we don't really care about the module itself here, we just care that it is in the correct place, and we're testing that here.

pedrobaeza: OK, if it's used only for that... I'm afraid one day it's used for other thing and then we lose the focus on this

Yajo: Don't worry. Actually I removed many tests when I deprecated doodba-scaffolding, and moved them to doodba-copier-template, now that we can have tests there.

So all I test here is that doodba puts things in its place, but in doodba-copier-template is where I test that part works fine.

pedrobaeza: OK, thanks for the clarifications.

PR #310, opened 2020-05-27
Revert "Update Python 3 version in v11 and v12"

Yajo:
This reverts commit dbaaa2782a2d00e093063ebee3478c1d4093def3 from https://github.com/Tecnativa/doodba/pull/290.

Also reverts v13 to oldest supported python version (3.6 per the docs).

@Tecnativa TT23435

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.7-stretch AS base
+FROM python:3.5-stretch AS base
 
 EXPOSE 8069 8072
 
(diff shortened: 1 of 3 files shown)


JordiBForgeFlow:
@Yajo is there any particular reason not revert to 3.6? We face some bugs with 3.5 that are fixed in 3.6

pedrobaeza:
Which bugs? The official Python version for v11 and v12 is P3.5.

Yajo:
The reason is that we chose to stick to the minimal supported version as stated officially, believing that it will be most supported.

FTR v13 was reverted to 3.6 for this same reason.

Another possibility would have been to update gevent, but that's completely away from the official requirements, and thus we didn't want to enter those grounds.

JordiBForgeFlow:
Ok. They are very project specific issues, and we fixed assuming 3.7. So no problem. We'll revert to using a 3.5 compatible fix.

pedrobaeza:
Is it worth to share that specific fixes?

Sorry anyway for changing this, but there's a problem with gevent + Odoo that we face it using upper versions (specially 3.7).

JordiBForgeFlow:
Not worth. These are scripts.

PR #313, opened 2020-06-04
Check pip packages consistency

Yajo:
Applying https://github.com/pypa/pip/issues/8104#issuecomment-638673688 everywhere to make sure builds do not include incompatible dependencies.


Yajo:
Problem didn't happen, so closing.

PR #324, opened 2020-09-30
Update ptvsd -> debugpy

joao-p-marques:
Seems like `debugpy` is the recommended library for debugging Python with VSCode, so it makes sense to make this transition.

https://code.visualstudio.com/docs/python/debugging#_debugging-by-attaching-over-a-network-connection 
https://github.com/microsoft/debugpy

This updates the container, but it is still needed to update the template under copier (see https://github.com/Tecnativa/doodba-copier-template/pull/127)

TT25875

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -20,8 +20,8 @@ ENV DB_FILTER=.* \
     OPENERP_SERVER=/opt/odoo/auto/odoo.conf \
     PATH="/home/odoo/.local/bin:$PATH" \
     PIP_NO_CACHE_DIR=0 \
-    PTVSD_ARGS="--host 0.0.0.0 --port 6899 --wait --multiprocess" \
-    PTVSD_ENABLE=0 \
+    DEBUGPY_ARGS="--listen 0.0.0.0:6899 --wait-for-client" \
+    DEBUGPY_ENABLE=0 \
     PUDB_RDB_HOST=0.0.0.0 \
     PUDB_RDB_PORT=6899 \
     PYTHONOPTIMIZE=1 \
@@ -82,7 +82,7 @@ RUN pip install \
         git-aggregator \
         pg_activity \
         plumbum \
-        ptvsd \
+        debugpy \
         pudb \
         watchdog \
         wdb \
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -20,6 +20,8 @@ ENV DB_FILTER=.* \
     OPENERP_SERVER=/opt/odoo/auto/odoo.conf \
     PATH="/home/odoo/.local/bin:$PATH" \
     PIP_NO_CACHE_DIR=0 \
+    PTVSD_ARGS="--host 0.0.0.0 --port 6899 --wait --multiprocess" \
+    PTVSD_ENABLE=0 \
     DEBUGPY_ARGS="--listen 0.0.0.0:6899 --wait-for-client" \
     DEBUGPY_ENABLE=0 \
     PUDB_RDB_HOST=0.0.0.0 \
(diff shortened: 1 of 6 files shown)


Yajo (approved):
It all looks good.

11.0.Dockerfile
```diff
@@ -20,8 +20,8 @@ ENV DB_FILTER=.* \
     OPENERP_SERVER=/opt/odoo/auto/odoo.conf \
     PATH="/home/odoo/.local/bin:$PATH" \
     PIP_NO_CACHE_DIR=0 \
-    PTVSD_ARGS="--host 0.0.0.0 --port 6899 --wait --multiprocess" \
-    PTVSD_ENABLE=0 \
+    DEBUGPY_ARGS="--listen 0.0.0.0:6899 --wait-for-client" \
```
Yajo, line 23: Thinking this a second time, we need a way to support backwards compatibility.

Otherwise, the only way to debug would be that the project is updated to the latest template and the latest image *or* based on an older image and template.

I'd prefer it if everything just works, so even if `PTVSD_` is no longer a cool prefix, I think we should still support it for a while.

For v14, after merging #323, we can support only `DEBUGPY_`, because there should exist no v14 projects out there.

Yajo (changes requested):

11.0.Dockerfile
joao-p-marques: Yes, that makes sense. I will make the changes.

joao-p-marques: Changes made. I also needed to change the copier template, as I changed the port where `debugpy` will run. See the related PR to see the changes

README.md
```diff
@@ -62,6 +62,7 @@ You can start working with this straight away with our [template][].
   - [`psql`](#psql)
   - [`inotify`](#inotify)
   - [`ptvsd`](#ptvsd)
```
Yajo, line 64: Please remove this section.

Although it still exists in old images, we don't want docs on deprecated stuff.

bin/direxec
```diff
@@ -53,6 +53,14 @@ if extra_command:
                 + os.environ["PTVSD_ARGS"].split()
                 + extra_command
             )
+        if os.environ["DEBUGPY_ENABLE"] == "1":
```
Yajo, line 56: We don't want both ptvsd and debugpy, right?

```suggestion
        elif os.environ["DEBUGPY_ENABLE"] == "1":
```

Also please add some deprecation notice when using ptvsd instead.

Yajo (changes requested):

joao-p-marques:
Changes made @Yajo

README.md
```diff
@@ -560,7 +563,7 @@ your project there is a `.vscode/launch.json` file with these minimal contents:
           "remoteRoot": "/opt/odoo"
         }
       ],
-      "port": 6899,
+      "port": 5678,
```
Yajo, line 566: This is wrong, isn't it?

bin/direxec
```diff
@@ -46,13 +46,27 @@ if extra_command:
                 # No DB exists, set initial language
                 extra_command += ["--load-language", os.environ["INITIAL_LANG"]]
         if os.environ["PTVSD_ENABLE"] == "1":
+            # Warn deprecation
+            logger.log(
+                WARNING,
+                "ptvsd has beed deprecated for python debugging. \
+                Please use debugpy (see https://github.com/Tecnativa/doodba#debugpy)",
```
Yajo, line 53: Just a little stylistic improvement:

```suggestion
                "ptvsd has beed deprecated for python debugging. "
                "Please use debugpy (see https://github.com/Tecnativa/doodba#debugpy)",
```

Yajo (changes requested):

README.md
joao-p-marques: Yes, ty! Fixed.

bin/direxec
joao-p-marques: Thank you

Yajo (approved):

PR #325, opened 2020-09-30
Use pytest for testing

Yajo:
@Tecnativa TT22772


Yajo:
Nah, it doesn't work. Sad :cry:

PR #326, opened 2020-09-30
Update CI to use Postgres 14

yajo:
@Tecnativa TT22772

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,13 +33,17 @@ jobs:
       fail-fast: false
       matrix:
         # Test modern Odoo versions with latest Postgres version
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -32,14 +32,15 @@ jobs:
       fail-fast: false
       matrix:
         # Test modern Odoo versions with latest Postgres version
-        pg_version: ["12"]
+        pg_version: ["13"]
         odoo_version:
```
joao-p-marques, line 36: Shouldn't you add version 14.0 here? Or does that go on a different PR?

Yajo: Yes, that's in #327

PR #332, opened 2020-10-20
Update 14.0 Dockerfile to use Python 3.8, as in Odoo.sh and runbots

joao-p-marques:
Seems like Odoo is using Python 3.8 for Runbots and Odoo.sh in v14: https://github.com/odoo/runbot/issues/414#issuecomment-710311902

TT26297

### DIFF
--- a/14.0.Dockerfile
+++ b/14.0.Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6-slim-buster AS base
+FROM python:3.8-slim-buster AS base
 
 EXPOSE 8069 8072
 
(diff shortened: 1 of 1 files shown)


Yajo (changes requested):
Please publish an announcement similar to https://github.com/Tecnativa/doodba/issues/67#issuecomment-634600294 (in that issue) before we merge.

Other than that, the patch is perfect :wink:

pedrobaeza:
It would be good to have the link to the comment that states about P3.8 in Odoo.sh and runbot.

joao-p-marques:
Done @Yajo and @pedrobaeza

Yajo (approved):

PR #333, opened 2020-10-21
Add Recordset debug collection support to Doodba under VSCode debugpy

gdgellatly:
Just a proposal at this stage, if desired we can make full PR with all Dockerfiles changed.

@Yajo what do you think? THis is what I was talking about on discord - I added screenshots as well.

### DIFF
--- a/13.0.Dockerfile
+++ b/13.0.Dockerfile
@@ -136,6 +136,7 @@ RUN build_deps=" \
         plumbum \
         ptvsd \
         debugpy \
+        pydevd-odoo \
         pudb \
         watchdog \
         wdb \


joao-p-marques (commented):
This seems like a good idea.
How well does `pydev-odoo` integrate with `debugpy`? It seems that `pydev` is bundled  with `debugpy`(https://github.com/microsoft/debugpy/tree/7a2891b95054f132435585f6ab5dedf66be4f5ab/src/debugpy/_vendored/pydevd) but have you tested the integration? Do we need to add configuration to debugpy to make it work?

Yajo:
I've tested this locally.

This is how inspecting looks like without the patch:

Peek 21-10-2020 09-31

And with it:

Peek 21-10-2020 09-41

I'm not very convinced. I'm not seeing any of the problems [you said](https://discord.com/channels/737652535149592587/765874049590362132/768262163978387466) without the patch, and I feel like `res.users(1,)` might be a better debug output than `[1] OdooBot` for `self` while debugging (the latter doesn't tell you which model it is, for example).

If you want to have this, all you need to do is add `pydev-odoo` to your `pip.txt` file, so it's not a big problem anyway.

I think maybe a reference in DCT docs for this should be enough for a "fix". I'd like to see what you think after reading this, @joao-p-marques and @pedrobaeza.

gdgellatly:
@Yajo Yes but you are looking at a recordset of 1 which the orm resolves. Try it on a sale order with a few lines and look at the order lines.

gdgellatly:
@joao-p-marques debugpy is kind of just the plumbing for pydevd and vscode. It doesn't actually do the debugging. PTVSD, Eclipse, pycharm all use pydevd for the heavy lifting. Over on the discord channel you can see screenshots, I have tested it in so far as I am using it right now and it seems to work perfectly.  Honestly it is a very simple module, it just checks if something is an instance of a Model, and if it is treats it like a collection rather than a single object.

Yajo:
OK, in multi-records I can see the benefits clearly:

Without patch:


With it:


I still don't like the fact that `self` becomes `[1, 2, 3]` instead of `stock.warehouse(1, 2, 3)`, but I guess that's a bug I can open in the project itself.

Seeing this.. I like it! @joao-p-marques @pedrobaeza WDYT?

pedrobaeza:
OK, the code itself seems simple for the library. Maybe we can host at OCA something similar (or this library itself with the permission of the author)

joao-p-marques:
Thanks for the explanation @gdgellatly 
It seems like a good idea then! :+1:

joao-p-marques (approved):

gdgellatly:
> OK, the code itself seems simple for the library. Maybe we can host at OCA something similar (or this library itself with the permission of the author)

@pedrobaeza I am kind of in agreement here. My first preference is obviously for the author to bring in to OCA. There does seem to be a language barrier.  Good IDE tools hosted in OCA is very much in line with our missions around quality. I wonder if we have any vietnamese community members that can assist us in the conversation.

pedrobaeza:
Maybe @jcdrubay can assist us on that.

Yajo:
I opened https://github.com/trinhanhngoc/pydevd-odoo/issues/1 regarding the specific problem found at https://github.com/Tecnativa/doodba/pull/333#issuecomment-713426058. And google translate does a good job bridging that language barrier already :blush: 

Well, that's out of scope I think. Regarding this PR, please @gdgellatly may you proceed with the full PR and take it out of draft?

gdgellatly:
@Yajo Interestingly it was from using google translate on that issue I figured we are better with a native speaker. Not convinced the translation is what was meant, or the question was understood. In any case responder was not maintainer.

I'll do PR now.

gdgellatly:
closed in favoor of #335

PR #334, opened 2020-10-21
Print debugpy logs to file

joao-p-marques:
They can further be analyzed by invoke to monitor if debug session is up

See https://github.com/Tecnativa/doodba-copier-template/pull/139

TT26189


Yajo:
Should we close now after last comments on https://github.com/Tecnativa/doodba-copier-template/pull/139?

Yajo (approved):

joao-p-marques:
Can we merge this? It is needed for https://github.com/Tecnativa/doodba-copier-template/pull/139 to work.

OTOH, I was thinking: If we print to `/opt/odoo/auto/debugpy-logs/`, as you said, the logs will persist on the container... However, they aren't really going to be necessary anyway, right? Besides, it introduces complexity in the `--wait` task, because you will then have to find which is the most recent file, and read that insted of always `debugpy.server-1.log`.

WDYT?

Yajo:
What I meant was that the logs should be somewhere around that folder, because `/opt/odoo/auto` is the only folder mapped for host-container shared files.

I didn't mean you should be using that path exactly, or that we should have a directory full of useless log files.

So yes, just use 1 file if possible please, it'll be easier :blush:

joao-p-marques:
Ok, thank you :smile: 
I ended up using the `/tmp` directory, which helps making sure we don't end up with log files we are never going to read... 
If we need for some reason, we can always override the `DEBUGPY_ARGS` env var in docker-compose.

Yajo:
Can this be closed now?

joao-p-marques:
Yes, we ended up following another strategy in https://github.com/Tecnativa/doodba-copier-template/pull/139

PR #361, opened 2020-12-16
Fix CI and add poetry

joao-p-marques:
Fixes tests error originated from using `cfssl` to test the usage of external dependencies. It was causing pip errors when installing for having a manifest with version 0.0.0 and only beta releases in pypi.

See the commit message for details


Yajo (approved):

Yajo:
I like the change, but it's still failing.
You might consider adding the poetry boilerplate to CI, to use it for tests.

joao-p-marques:
Thanks. It is failing, but for different reasons... I tried to reference a file in master before it arrived there :sweat_smile: 
I think the last commit should fix it.
I will work on using poetry in the tests in the meantime.

joao-p-marques:
@Yajo changes made and tests passing :heavy_check_mark:

Yajo (approved):

PR #365, opened 2020-12-23
image: enable defining postgres client version on build

ap-wtioit:
When using pg_dump in the odoo container we had the problem that we couldn't restore the dumps on our main postgres server because they where using a different pg_dump version than the pg_restore outside of the container.

With this change the version of pg_dump, psql and pg_restore inside of the odoo container can be set by adding the build parameter `DB_VERSION: ${DB_VERSION}` to use the DB_VERSION from .env or `DB_VERSION: 10` to specify a specific version for the odoo containers postgres-client package.

Unfortunately just specifying postgres-client-10 in dependencies/apt.txt did install psql version 10 but did not set as the default psql in the odoo user's path (that was still version 13)

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -55,7 +55,7 @@ RUN apt-get -qq update \
     && curl https://bootstrap.pypa.io/get-pip.py | python3 /dev/stdin \
     && curl -sL https://deb.nodesource.com/setup_6.x | bash - \
     && apt-get update \
-    && apt-get install -yqq --no-install-recommends nodejs postgresql-client \
+    && apt-get install -yqq --no-install-recommends nodejs \
     && curl -SLo wkhtmltox.deb https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/${WKHTMLTOPDF_VERSION}/wkhtmltox_${WKHTMLTOPDF_VERSION}-1.stretch_amd64.deb \
     && echo "${WKHTMLTOPDF_CHECKSUM}  wkhtmltox.deb" | sha256sum -c - \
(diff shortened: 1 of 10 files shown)


pedrobaeza:
Why aren't you doing the dump with the db container?

ap-wtioit:
The dump in the db container is unfortunately done by the the "wrong" user and needs chowning for protection / use on other systems.

Also with performing pg_dump inside of odoo container we can run a full backup inside of the odoo container (consisting of the filestore db) its easier to coordinate the dump and filestore for a certain point in time.

pedrobaeza:
OK, thanks for the explanations.

11.0.Dockerfile
```diff
@@ -221,6 +221,7 @@ ONBUILD RUN mkdir -p /opt/odoo/custom/ssh \
             && ln -s /opt/odoo/custom/ssh ~root/.ssh \
             && chmod -R u=rwX,go= /opt/odoo/custom/ssh \
             && sync
+ONBUILD ARG DB_VERSION=None
```
Yajo, line 224: This is a bit misleading. I think the best is to make it default to "latest" instead. You'll have to fix the other dockerfiles and the build script too.

Yajo (changes requested):
Just a small comment, but other than that this PR looks very good. Thanks.

I think this deserves a complementary PR for the template.

Yajo (approved):

Yajo:
@pedrobaeza This looks good, has tests and they pass, so it should be safe to merge. But how I'm gonna be out for some days, I leave it up to you if you want to merge now or wait.

pedrobaeza:
Go ahead. Technically you have the word.

PR #370, opened 2021-01-12
Add option to have multiple addons versions defined in different environments

joao-p-marques:
Checking for duplicates in `doodbalib` was impeding having the same addon defined more than once in `addons.yaml`, even if it was for different environments.
This moves that check to the entrypoint, to be executed only when symlinking the addons.
Also add tests for that situation

@Tecnativa
TT27711

### DIFF
--- a/lib/doodbalib/__init__.py
+++ b/lib/doodbalib/__init__.py
@@ -159,11 +159,12 @@ def addons_config(filtered=True, strict=False):
             continue
         repos.discard(CORE)
         # Other addons fall in between
-        if len(repos) != 1:
+        if filtered and len(repos) != 1:
             raise AddonsConfigError(
                 u"Addon {} defined in several repos {}".format(addon, repos)
             )
-        yield addon, repos.pop()
+        for repo in repos:
+            yield addon, repo
 
 
 try:
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -327,6 +327,35 @@ class ScaffoldingCase(unittest.TestCase):
                 ("test", "-d", "auto/addons/crm/migrations"),
             )
 
+    # HACK https://github.com/itpp-labs/misc-addons/issues/1014
+    # TODO Remove decorator
+    @prerelease_skip
+    def test_addons_env_double(self):
+        """Test double addon reference in addons.yaml"""
+        common_tests = (
+            ("test", "-d", "custom/src/rma-old/rma"),
+            ("test", "!", "-d", "custom/src/rma-old/rma_sale"),
+            ("test", "-d", "custom/src/rma-new/rma"),
+            ("test", "!", "-d", "custom/src/rma-new/rma_sale"),
+        )
+        # Old versions are skiped because they don't support __manifest__.py,
+        # and the test is hacking ODOO_VERSION to pin a commit
+        for sub_env in matrix(odoo_skip={"7.0", "8.0", "9.0"}):
+            self.compose_test(
+                join(SCAFFOLDINGS_DIR, "addons_env_double"),
+                dict(sub_env, DOODBA_ENVIRONMENT="test"),
+                *common_tests,
+                # Check version is 12.0.1.6.1
+                ("grep", "-q", "12.0.1.6.1", "auto/addons/rma/__manifest__.py"),
+            )
+            self.compose_test(
+                join(SCAFFOLDINGS_DIR, "addons_env_double"),
+                dict(sub_env, DOODBA_ENVIRONMENT="prod"),
+                *common_tests,
+                # Check version is 12.0.2.0.0
+                ("grep", "-q", "12.0.2.0.0", "auto/addons/rma/__manifest__.py"),
+            )
+
     def test_dotd(self):
         """Test environment with common ``*.d`` directories."""
         for sub_env in matrix():
--- a/tests/scaffoldings/addons_env_double/custom/src/repos.yaml
+++ b/tests/scaffoldings/addons_env_double/custom/src/repos.yaml
@@ -1,0 +1,25 @@
+# Odoo is always required
+./odoo:
+  defaults:
+    # Shallow repositores are faster & thinner
+    depth: $DEPTH_DEFAULT
+  remotes:
+    ocb: https://github.com/OCA/OCB.git
+    odoo: https://github.com/odoo/odoo.git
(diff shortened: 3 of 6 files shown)


entrypoint.d/40-addons-link
```diff
@@ -3,15 +3,31 @@
...
+addons_list = addons_config()
+for addon, repo in addons_list:
+    # Check for doubled
+    occurrences = [i for i in addons_list if i[0] == addon]
+    if len(occurrences) != 1:
+        raise AddonsConfigError(
+            u"Addon {} defined in several repos {}".format(
+                addon, [i[1] for i in occurrences]
+            )
+        )
```
Yajo, line 30: You probably don't need this. See below.

lib/doodbalib/__init__.py
```diff
@@ -159,10 +159,6 @@ def addons_config(filtered=True, strict=False):
             continue
         repos.discard(CORE)
         # Other addons fall in between
-        if len(repos) != 1:
```
Yajo, line 162: The correct fix would be probably:

```py
if filtered and len(repos) != 1:
```

Because it will usually run with `filtered=True`, except when building.

lib/doodbalib/__init__.py
```diff
@@ -159,10 +159,6 @@ def addons_config(filtered=True, strict=False):
...
-            raise AddonsConfigError(
-                u"Addon {} defined in several repos {}".format(addon, repos)
-            )
         yield addon, repos.pop()
```
Yajo, line 162: Since there could be now several `repos`, you'd need a loop here too.

tests/scaffoldings/addons_env_double/custom/src/repos.yaml
```diff
@@ -0,0 +1,20 @@
...
+  target: ocb $ODOO_VERSION
+  merges:
+    - ocb $ODOO_VERSION
+
+rma-old:
+  defaults:
+    depth: $DEPTH_MERGE
+  remotes:
+    origin: https://github.com/OCA/rma.git
+  target: origin 242b7e21a174e48f713459f836794560e68290de # should be before commit 96d2081987292ae02efda9826cbe5ba1ad84d6c3
```
Yajo, line 18: The comment is not much helpful... Also, `target` must be a branch name:

```suggestion
  target: origin _merged_branch
```

tests/scaffoldings/addons_env_double/custom/src/repos.yaml
```diff
@@ -0,0 +1,20 @@
...
+  merges:
+    - origin $ODOO_VERSION
```
Yajo, line 20:
```suggestion
    - origin 242b7e21a174e48f713459f836794560e68290de # rma v12.0.1.6.1
```

tests/scaffoldings/addons_env_double/custom/src/repos.yaml
```diff
@@ -0,0 +1,20 @@
...
+  remotes:
+    ocb: https://github.com/OCA/OCB.git
+    odoo: https://github.com/odoo/odoo.git
+  target: ocb $ODOO_VERSION
+  merges:
+    - ocb $ODOO_VERSION
+
+rma-old:
+  defaults:
+    depth: $DEPTH_MERGE
```
Yajo, line 15: You can't use this option with direct commits:

```suggestion
```

tests/__init__.py
```diff
@@ -327,6 +327,41 @@ def test_addons_env(self):
...
+                join(SCAFFOLDINGS_DIR, "addons_env_double"),
+                sub_env,
+                # check module from custom repo pattern
+                ("test", "-d", "custom/src/rma"),
+                ("test", "-d", "custom/src/rma/rma"),
+                ("test", "!", "-d", "custom/src/rma-old"),
+                ("test", "!", "-d", "custom/src/rma-old/rma"),
+                ("test", "-e", "auto/addons/rma"),
+                # Check version is 12.0.2.0.0
+                ("grep", "-Rq", "12.0.2.0.0", "auto/addons/rma/rma"),
```
Yajo, line 362: This test will fail the day 12.0.2.0.0.1 is released. Also on non-v12 builds. Make it more resilient please. Probably, you need to pin the commit for the normal version too in repos.yaml.

tests/scaffoldings/addons_env_double/prod.yaml
```diff
@@ -0,0 +1,29 @@
...
+      args:
+        COMPILE: "false"
+        ODOO_VERSION: $ODOO_MINOR
+        WITHOUT_DEMO: "false"
+    tty: true
+    depends_on:
+      - db
+    environment:
+      PYTHONOPTIMIZE: ""
+      UNACCENT: "false"
```
Yajo, line 15: I don't think you really need 2 separate docker-compose files if you add this here:

```suggestion
    environment:
      DOODBA_ENVIRONMENT: "$DOODBA_ENVIRONMENT"
      PYTHONOPTIMIZE: ""
      UNACCENT: "false"
```

Of course, you must add that env variable to the test execution env too.

Yajo (changes requested):

joao-p-marques:
Thank you @Yajo 
Changes made

Yajo (approved):

tests/__init__.py
```diff
@@ -327,6 +327,33 @@ def test_addons_env(self):
...
+    @prerelease_skip
+    def test_addons_env_double(self):
+        """Test double addon reference in addons.yaml"""
+        common_tests = (
+            ("test", "-d", "custom/src/rma-old/rma"),
+            ("test", "!", "-d", "custom/src/rma-old/rma_sale"),
+            ("test", "-d", "custom/src/rma-new/rma"),
+            ("test", "!", "-d", "custom/src/rma-new/rma_sale"),
+        )
+        for sub_env in matrix():
```
Yajo, line 341: There must be some check about the `__manifest__.py` stuff that I forgot about: https://github.com/Tecnativa/doodba/pull/370/checks?check_run_id=1694934996#step:8:192

Please skip <10.0 versions

Yajo (changes requested):

tests/__init__.py
joao-p-marques: Ok, I was thinking the same :smile: :+1: Thanks

Yajo (approved):

PR #396, opened 2021-03-08
[FIX] dependencies: update pip install url and replace yanked aloha_world gem

ap-wtioit:
During our nightly build the builds for 11.0 and 12.0 failed with the following message:
```
Hi there!
The URL you are using to fetch this script has changed, and this one will no
longer work. Please use get-pip.py from the following URL instead:
    https://bootstrap.pypa.io/pip/3.5/get-pip.py
Sorry if this change causes any inconvenience for you!
We don't have a good mechanism to make more gradual changes here, and this
renaming is a part of an effort to make it easier to us to update these
scripts, when there's a pip release. It's also essential for improving how we
handle the `get-pip.py` scripts, when pip drops support for a Python minor
version.
There are no more renames/URL changes planned, and we don't expect that a need
would arise to do this again in the near future.
Thanks for understanding!
- Pradyun, on behalf of the volunteers who maintain pip.
```

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -52,7 +52,7 @@ RUN apt-get -qq update \
         zlibc \
     && echo 'deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main' >> /etc/apt/sources.list.d/postgresql.list \
     && curl -SL https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
-    && curl https://bootstrap.pypa.io/3.5/get-pip.py | python3 /dev/stdin \
+    && curl https://bootstrap.pypa.io/pip/3.5/get-pip.py | python3 /dev/stdin \
     && curl -sL https://deb.nodesource.com/setup_6.x | bash - \
     && apt-get update \
     && apt-get install -yqq --no-install-recommends nodejs \
--- a/12.0.Dockerfile
+++ b/12.0.Dockerfile
@@ -52,7 +52,7 @@ RUN apt-get -qq update \
         zlibc \
     && echo 'deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main' >> /etc/apt/sources.list.d/postgresql.list \
     && curl -SL https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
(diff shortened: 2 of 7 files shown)


joao-p-marques:
Hi, thanks for the contribution! :smiley: 

It seems like this would also be necessary for lower versions: https://github.com/Tecnativa/doodba/pull/396/checks?check_run_id=2054401931#step:7:665
Could you please make the same change for the other images?

joao-p-marques (approved):

Yajo (changes requested):
Could you please check the errors like this: https://github.com/Tecnativa/doodba/pull/396/checks?check_run_id=2054482698#step:8:5008

```
  File "/usr/local/bin/autoaggregate", line 66, in aggregate
    os.chown(os.path.join(root, target), UID, GID)
FileNotFoundError: [Errno 2] No such file or directory: '/opt/odoo/custom/src/dummy_repo/dummy_link'
```

ap-wtioit:
@Yajo i will check this. i suspect it's not caused by this merge request (but i'll verify / fix it)

ap-wtioit:
@Yajo tests are fixed by replacing test dependency [aloha_world](https://rubygems.org/gems/aloha_world) with [hello-world](https://rubygems.org/gems/hello-world)

ap-wtioit:
The dummy_link error:
```
  File "/usr/local/bin/autoaggregate", line 66, in aggregate
    os.chown(os.path.join(root, target), UID, GID)
FileNotFoundError: [Errno 2] No such file or directory: '/opt/odoo/custom/src/dummy_repo/dummy_link'
```
doesn't seem to affect the tests and seems to have been introduced in https://github.com/Tecnativa/doodba/commit/3f5644dd3e3892ffbe4e854d634f4012b2a59b71 when the dummy_addon was moved from tests/scaffoldings/dotd/custom/src/dummy_repo to tests/scaffoldings/dotd/custom/src/other-doodba/odoo/src/private/dummy_addon without changing the dummy_link addon. 

It doesn't break the tests though.

Should this give a test error? Or is it a test that aggregation still works when there is a broken symlink? (Does aggregation still work when there is a broken symlink?)

However i feel this is out of scope of this pull request (I'll open a new one for that)

Yajo (approved):

PR #399, opened 2021-03-08
[FIX] autoaggrate: do not chown symlinks

ap-wtioit:
when chowning symlinks, os.chown modifies the target instead of the symlink
as the target (if under SRC_DIR) should be chowned separately it's ok to skip
chowning symlinks. this prevents unclear error messages where it looks like the
symlink does not exist (when in fact it's the target file) and gives the same
result as if we are chowning targets via symlinks twice.


ap-wtioit:
This removes the error mentioned in https://github.com/Tecnativa/doodba/pull/396
```
  File "/usr/local/bin/autoaggregate", line 66, in aggregate
    os.chown(os.path.join(root, target), UID, GID)
FileNotFoundError: [Errno 2] No such file or directory: '/opt/odoo/custom/src/dummy_repo/dummy_link'
```

It should not have any effect besides getting rid of FileNotFoundErrors for symlinks pointing to nowhere in SRC_DIR

joao-p-marques (approved):

ap-wtioit:
Well i cannot think of a good test for that:
* It just suppresses the log message in autoaggregate, as autoaggregate seems only be called/be designed for calling inside the build process the build process is is not really testable inside a docker-compose run (without a lot of docker in docker magic)
** autoaggregate needs root permissions which are not available to odoo after the build
* For the filesystem i could only test the permissions of files inside the src directory, but as permissions are the same with and without the patch the test would not fail before the patch
* The other effect i tried to achieve (the maintainers not reporting this error message on my merge requests) seems also not automatically testable

Can you think of a test that is runnable inside a docker-compose run (our test framework), fails before the patch and succeeds after the patch?

Yajo:
I was thinking of this part:

> when chowning symlinks, os.chown modifies the target instead of the symlink

You could produce a good symlink and then assert that its target is not chowned.

ap-wtioit:
Yes, that would work (pointing the symlink outside of SRC_DIR), i'll add a test, thanks

tests/scaffoldings/symlinks/Dockerfile
```diff
@@ -0,0 +1,2 @@
+ARG ODOO_VERSION
+FROM registry.wt-io-it.at/local:ci-local-${ODOO_VERSION}-onbuild
```
joao-p-marques, line 2: Can you change this to:
```suggestion
FROM tecnativa/doodba:${ODOO_VERSION}-onbuild
```
which is what we use on the rest of the tests? 
Or is there a need to use this specific image in your test?

joao-p-marques (commented):
Thanks for the tests! 🙂 :+1:

tests/scaffoldings/symlinks/Dockerfile
ap-wtioit: sorry, yes, will fix that, this is an artifact of my local tests

ap-wtioit:
Current error https://pipelines.actions.githubusercontent.com/RaPZd2A6k1gqIXsl1jltoa1XTQGB7pDVtOtE10iOBn29R7SodM/_apis/pipelines/1/runs/161/signedlogcontent/17?urlExpires=2021-03-10T13%3A13%3A21.8974373Z&urlSigningMethod=HMACV1&urlSignature=u%2BabdFgKNIK66qAGYkWILQH2gBsjVrZdUiSb47G17qk%3D:
```
2021-03-10T12:49:00.2556373Z Requirement already satisfied: pg_activity in /usr/local/lib/python3.6/site-packages (2.1.0)
2021-03-10T12:49:00.2624676Z Requirement already satisfied: humanize in /usr/local/lib/python3.6/site-packages (from pg_activity) (3.2.0)
2021-03-10T12:49:00.2634221Z Requirement already satisfied: attrs in /usr/local/lib/python3.6/site-packages (from pg_activity) (20.3.0)
2021-03-10T12:49:00.2645889Z Requirement already satisfied: psutil>=2.0.0 in /usr/local/lib/python3.6/site-packages (from pg_activity) (5.6.6)
2021-03-10T12:49:00.2659923Z Requirement already satisfied: blessed in /usr/local/lib/python3.6/site-packages (from pg_activity) (1.18.0)
2021-03-10T12:49:00.2972656Z Requirement already satisfied: wcwidth>=0.1.4 in /usr/local/lib/python3.6/site-packages (from blessed->pg_activity) (0.2.5)
2021-03-10T12:49:00.2982726Z Requirement already satisfied: six>=1.9.0 in /usr/local/lib/python3.6/site-packages (from blessed->pg_activity) (1.15.0)
2021-03-10T12:49:00.3056485Z Requirement already satisfied: setuptools in /usr/local/lib/python3.6/site-packages (from humanize->pg_activity) (53.0.0)
2021-03-10T12:49:00.8532767Z + pg_activity --version
2021-03-10T12:49:00.9607897Z /usr/local/lib/python3.6/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
2021-03-10T12:49:00.9609703Z   """)
2021-03-10T12:49:00.9614899Z Traceback (most recent call last):
2021-03-10T12:49:00.9616985Z   File "/usr/local/bin/pg_activity", line 36, in <module>
2021-03-10T12:49:00.9617718Z     from psycopg2.errors import OperationalError
2021-03-10T12:49:00.9618900Z ModuleNotFoundError: No module named 'psycopg2.errors'
2021-03-10T12:49:00.9732314Z Traceback (most recent call last):
2021-03-10T12:49:00.9733189Z   File "/opt/odoo/common/entrypoint", line 30, in <module>
2021-03-10T12:49:00.9733836Z     subprocess.check_call(command)
2021-03-10T12:49:00.9734580Z   File "/usr/local/lib/python3.6/subprocess.py", line 311, in check_call
2021-03-10T12:49:00.9735371Z     raise CalledProcessError(retcode, cmd)
2021-03-10T12:49:00.9737036Z subprocess.CalledProcessError: Command '/opt/odoo/custom/entrypoint.d/pip-install-user' returned non-zero exit status 1.
```

pg_activity 2.1.0 (release on Monday) is not compatible with installed psycopq2 package (pg_activity 2.0.3 was)

Yajo (approved):
Nice, thanks!

joao-p-marques (approved):
Thanks a lot for your hard work and patience with this fix @ap-wtioit :+1: 
Your contribution is greatly appreciated :smiley:

ap-wtioit:
thanks guys, for taking care of this great project and for the feedback you provide, i appreciate it a lot.

PR #407, opened 2021-03-30
Add preparedb script

joao-p-marques:
Prepopulates DB with helpful simple data
Add tests

TT27635

### DIFF
--- a/bin/preparedb
+++ b/bin/preparedb
@@ -1,0 +1,18 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Script to prepare the database with initial data
+
+import click
+import click_odoo
+
+
+@click.command()
+@click_odoo.env_options(default_log_level="info", database_must_exist=True)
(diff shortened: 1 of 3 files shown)


bin/preparedb
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env click-odoo
```
Yajo, line 1: I prefer it if you set here:

```suggestion
#!/usr/bin/env python
```

Then, inside python, you import click_odoo.

Take a look into any script from click-odoo-contrib, such as https://github.com/acsone/click-odoo-contrib/blob/4fa3b8f0b4ad6a12042fffde8a54150accf79f5c/click_odoo_contrib/dropdb.py

This way, the file will be detected as python and used by linters, formatters, syntax highlighters, etc.

Yajo (changes requested):
Just one change.

Yajo (approved):

PR #413, opened 2021-04-12
Improve build performance

ap-wtioit:
On some of our build systems (2/3) we seem to hit this issue:
https://github.com/docker/for-linux/issues/388

With this merge request:
* The number of needed `chown` and `chmod` operations is greatly reduced so we can finish our CI Tests on those systems within the timeout (1h)
* autoaggregate now checks out files as odoo user right away (with the correct umask as default)
* python pre compile in /opt/odoo/custom/src is now performed as odoo user to reduce the need of chowning the python cache after compilation

* /opt/odoo/custom/src now belongs to odoo:odoo instead of root:odoo (this seemed to be more correct for development where odoos UID and GID are mapped to the developers UID and GID)

### DIFF
--- a/13.0.Dockerfile
+++ b/13.0.Dockerfile
@@ -162,6 +162,18 @@ LABEL org.label-schema.schema-version="$VERSION" \
 # Onbuild version, with all the magic
 FROM base AS onbuild
 
+# Enable setting custom uids for odoo user during build of scaffolds
+ONBUILD ARG UID=1000
+ONBUILD ARG GID=1000
+
+# Enable Odoo user and filestore
+ONBUILD RUN groupadd -g $GID odoo -o \
+    && useradd -l -md /home/odoo -s /bin/false -u $UID -g $GID odoo \
+    && mkdir -p /var/lib/odoo \
+    && chown -R odoo:odoo /var/lib/odoo /qa/artifacts \
+    && chmod a=rwX /qa/artifacts \
+    && sync
+
 # Subimage triggers
 ONBUILD ENTRYPOINT ["/opt/odoo/common/entrypoint"]
 ONBUILD CMD ["/usr/local/bin/odoo"]
@@ -210,19 +222,7 @@ ONBUILD ENV ADMIN_PASSWORD="$ADMIN_PASSWORD" \
             EMAIL_FROM="$EMAIL_FROM" \
             WITHOUT_DEMO="$WITHOUT_DEMO"
 ONBUILD ARG LOCAL_CUSTOM_DIR=./custom
-ONBUILD COPY $LOCAL_CUSTOM_DIR /opt/odoo/custom
-
-# Enable setting custom uids for odoo user during build of scaffolds
-ONBUILD ARG UID=1000
-ONBUILD ARG GID=1000
-
-# Enable Odoo user and filestore
-ONBUILD RUN groupadd -g $GID odoo -o \
-    && useradd -l -md /home/odoo -s /bin/false -u $UID -g $GID odoo \
-    && mkdir -p /var/lib/odoo \
-    && chown -R odoo:odoo /var/lib/odoo /qa/artifacts \
-    && chmod a=rwX /qa/artifacts \
-    && sync
+ONBUILD COPY --chown=root:odoo $LOCAL_CUSTOM_DIR /opt/odoo/custom
 
 # https://docs.python.org/3/library/logging.html#levels
 ONBUILD ARG LOG_LEVEL=INFO
--- a/bin/autoaggregate
+++ b/bin/autoaggregate
@@ -10,7 +10,6 @@ from doodbalib import (
     ADDONS_YAML,
     AUTO_REPOS_YAML,
     CORE,
-    LOG_LEVELS,
     ODOO_DIR,
     PRIVATE,
     REPOS_YAML,
@@ -18,7 +17,9 @@ from doodbalib import (
     logger,
 )
 
-UMASK = os.environ.get("UMASK")
+# if the umask matches the `chmod -R u+rwX,g+rX-w,o= /opt/odoo` command the build is faster as we don't need to fix as
+# many permissions after auto aggregation
+UMASK = os.environ.get("UMASK") or "0027"
 UID = int(os.environ.get("UID") or -1)
 GID = int(os.environ.get("GID") or -1)
 DEFAULT_REPO_PATTERN = os.environ.get("DEFAULT_REPO_PATTERN")
@@ -33,47 +34,58 @@ def aggregate(config):
         Path where to find the ``repos.yaml`` file.
     """
     logger.info("Running gitaggregate with %s", config)
-    old_umask = None
-    try:
-        # Download git code with the specified umask, if any
-        if UMASK:
-            old_umask = os.umask(int(UMASK))
-        check_call(
-            [
-                "gitaggregate",
-                "--expand-env",
-                "--config",
-                config,
-                "--log-level",
-                log_level,
-                "--jobs",
-                str(cpu_count() or 1),
-                "aggregate",
-            ],
-            cwd=SRC_DIR,
-            stderr=sys.stderr,
-            stdout=sys.stdout,
-        )
-    finally:
-        # Restore umask, if changed
-        if old_umask is not None:
-            os.umask(old_umask)
-        # Chown recursively, if UID or GID are specified
-        if ~UID or ~GID:
-            for root, dirs, files in os.walk(SRC_DIR):
-                for target in dirs + files:
-                    try:
-                        target_path = os.path.join(root, target)
-                        if not os.path.islink(target_path):
-                            # if target_path is a link we would only chown the
-                            # target, as the target should be chowned separately
-                            # we can skip all symlinks and prevent showing
-                            # unclear errors
-                            os.chown(target_path, UID, GID)
-                    except Exception:
-                        logger.debug(
-                            "Error trying to chown on file. Skipping...", exc_info=True
-                        )
+
+    def pre_exec_umask():
+        # Download git code with the specified umask, if set, otherwise use "0027"
+        os.umask(int(UMASK, 8))
+
+    pre_execs = [pre_exec_umask]
+
+    def pre_exec():
+        for _exec in pre_execs:
+            try:
+                _exec()
+            except Exception as e:
+                logger.error("Error in %s: %s" % (_exec, e))
+                logger.exception(e)
+                raise
+
+    if ~GID:
+
+        def pre_exec_gid():
+            # execute git with GID
+            os.setgid(GID)
+
+        pre_execs.append(pre_exec_gid)
+
+    if ~UID:
+
+        def pre_exec_uid():
+            # execute git with UID
+            os.setuid(UID)
+            # set odoo home directory
+            # (git checks if user has a config in $HOME, and we cannot read /root as odoo user)
+            os.environ["HOME"] = "/home/odoo"
+
+        pre_execs.append(pre_exec_uid)
+
+    check_call(
+        [
+            "gitaggregate",
+            "--expand-env",
+            "--config",
+            config,
+            "--log-level",
+            log_level,
+            "--jobs",
+            str(cpu_count() or 1),
+            "aggregate",
+        ],
+        cwd=SRC_DIR,
+        stderr=sys.stderr,
+        stdout=sys.stdout,
+        preexec_fn=pre_exec,
+    )
 
 
 def origin_for(
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -615,6 +615,30 @@ class ScaffoldingCase(unittest.TestCase):
                 ),
             )
 
+    def test_repo_merge(self):
+        symlink_dir = join(SCAFFOLDINGS_DIR, "repo_merge")
+        for sub_env in matrix():
+            self.compose_test(
+                symlink_dir,
+                sub_env,
+                (
+                    "git",
+                    "--git-dir=/opt/odoo/custom/src/odoo/.git",
+                    "--no-pager",
+                    "log",
+                    "-n",
+                    "2",
+                ),
+                # make sure the git log contains a merge commit signed by $GIT_AUTHOR_NAME (otherwise ff was enabled
+                # and we did not test the merge commits)
+                (
+                    "bash",
+                    "-c",
+                    "git --git-dir=/opt/odoo/custom/src/odoo/.git log -n 1"
+                    " | grep 'docker-odoo <https://hub.docker.com/r/tecnativa/odoo>'",
+                ),
+            )
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -640,6 +640,39 @@ class ScaffoldingCase(unittest.TestCase):
                 ),
             )
 
+    def test_repo_merge_aggregate_permissions(self):
+        symlink_dir = join(SCAFFOLDINGS_DIR, "repo_merge")
+        for sub_env in matrix():
+            self.compose_test(
+                symlink_dir,
+                dict(
+                    sub_env,
+                    COMPOSE_FILE="setup-devel.yaml",
+                    UID=str(os.getuid()),
+                    GID=str(os.getgid()),
+                ),
+                # prepare repos.yaml to be non fast forward
+                ("/opt/odoo/custom/build.d/099-git_merge_no_ff",),
+                # autoaggregate as odoo:odoo to check if merges also work
+                ("autoaggregate",),
+                (
+                    "git",
+                    "--git-dir=/opt/odoo/custom/src/odoo/.git",
+                    "--no-pager",
+                    "log",
+                    "-n",
+                    "2",
+                ),
+                # make sure the git log contains a merge commit signed by $GIT_AUTHOR_NAME (otherwise ff was enabled
+                # and we did not test the merge commits)
+                (
+                    "bash",
+                    "-c",
+                    "git --git-dir=/opt/odoo/custom/src/odoo/.git log -n 1"
+                    " | grep 'docker-odoo <https://hub.docker.com/r/tecnativa/odoo>'",
+                ),
+            )
+
     def test_aggregate_permissions(self):
         symlink_dir = join(SCAFFOLDINGS_DIR, "aggregate_permissions")
         for sub_env in matrix():
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -5,6 +5,7 @@
 Each test must be a valid docker-compose.yaml file with a ``odoo`` service.
 """
 import logging
+import os
 import unittest
 from itertools import product
 from os import environ
@@ -637,6 +638,18 @@ class ScaffoldingCase(unittest.TestCase):
                     "git --git-dir=/opt/odoo/custom/src/odoo/.git log -n 1"
                     " | grep 'docker-odoo <https://hub.docker.com/r/tecnativa/odoo>'",
                 ),
+            )
+
+    def test_aggregate_permissions(self):
+        symlink_dir = join(SCAFFOLDINGS_DIR, "aggregate_permissions")
+        for sub_env in matrix():
+            self.compose_test(
+                symlink_dir,
+                dict(sub_env, UID=str(os.getuid()), GID=str(os.getgid())),
+                ("autoaggregate",),
+                # test that permissions are set in a way that enables a second autoaggregation after the first one
+                # e.g. when used in dev we update the source code sometimes/daily
+                ("autoaggregate",),
             )
 
 
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -157,6 +157,18 @@ LABEL org.label-schema.schema-version="$VERSION" \
 # Onbuild version, with all the magic
 FROM base AS onbuild
 
+# Enable setting custom uids for odoo user during build of scaffolds
+ONBUILD ARG UID=1000
+ONBUILD ARG GID=1000
+
+# Enable Odoo user and filestore
+ONBUILD RUN groupadd -g $GID odoo -o \
+    && useradd -l -md /home/odoo -s /bin/false -u $UID -g $GID odoo \
+    && mkdir -p /var/lib/odoo \
+    && chown -R odoo:odoo /var/lib/odoo /qa/artifacts \
+    && chmod a=rwX /qa/artifacts \
+    && sync
+
 # Subimage triggers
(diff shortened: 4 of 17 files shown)


ap-wtioit:
Note: test errors seem unrelated to merge request and could not be reproduced locally:
```
#20 3.406 Requirement already satisfied: click in /usr/local/lib/python3.5/site-packages
#20 3.430 Collecting coverage
#20 4.008   Downloading https://files.pythonhosted.org/packages/18/97/fda94f1e78b508dd1a493106c236076c4c82a9925c9347211afc4d6bd4e0/coverage-5.5-cp35-cp35m-manylinux1_x86_64.whl (240kB)
#20 4.087 Collecting flake8
#20 4.190   Downloading https://files.pythonhosted.org/packages/2a/cb/cd92e789442e234b8701bf6e886a55fbc83b7fd6e529b047e20b9cf196e8/flake8-3.9.0-py2.py3-none-any.whl (73kB)
#20 4.252 Collecting pylint-odoo
#20 4.321   Downloading https://files.pythonhosted.org/packages/74/ae/9cc3d387291e1e2866e0af906e1b76e766885867fe765a3f296f5879fd42/pylint-odoo-3.7.1.tar.gz (69kB)
#20 5.055     Complete output from command python setup.py egg_info:
#20 5.055     Couldn't find index page for 'pbr' (maybe misspelled?)
#20 5.055     No local packages or working download links found for pbr
#20 5.055     Traceback (most recent call last):
#20 5.055       File "<string>", line 1, in <module>
#20 5.055       File "/tmp/pip-build-9p5nu15_/pylint-odoo/setup.py", line 8, in <module>
#20 5.055         package_data={'': ['*.yaml']})
#20 5.055       File "/usr/local/lib/python3.5/distutils/core.py", line 108, in setup
#20 5.055         _setup_distribution = dist = klass(attrs)
#20 5.055       File "/qa/venv/lib/python3.5/site-packages/setuptools/dist.py", line 315, in __init__
#20 5.055         self.fetch_build_eggs(attrs['setup_requires'])
#20 5.055       File "/qa/venv/lib/python3.5/site-packages/setuptools/dist.py", line 361, in fetch_build_eggs
#20 5.055         replace_conflicting=True,
#20 5.055       File "/qa/venv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 850, in resolve
#20 5.055         dist = best[req.key] = env.best_match(req, ws, installer)
#20 5.055       File "/qa/venv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 1122, in best_match
#20 5.055         return self.obtain(req, installer)
#20 5.055       File "/qa/venv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 1134, in obtain
#20 5.055         return installer(requirement)
#20 5.055       File "/qa/venv/lib/python3.5/site-packages/setuptools/dist.py", line 429, in fetch_build_egg
#20 5.055         return cmd.easy_install(req)
#20 5.055       File "/qa/venv/lib/python3.5/site-packages/setuptools/command/easy_install.py", line 659, in easy_install
#20 5.055         raise DistutilsError(msg)
#20 5.055     distutils.errors.DistutilsError: Could not find suitable distribution for Requirement.parse('pbr')
```

ap-wtioit:
TODOs @ap-wtioit: there seems to be issues with merges defined in scaffoldings

- [ ] add test for merges in scaffoldings
- [ ] fix merges in scaffoldings

joao-p-marques:
@ap-wtioit thanks for your contribution once more :smiley: 

> Note: test errors seem unrelated to merge request and could not be reproduced locally:

Indeed, it is an unrelated error. Taking a look at some [similar errors](https://github.com/googleapis/google-cloud-python/issues/3757#issuecomment-321029449), it seems to be an issue with the Pypi mirrors or firewall restrictions in Github Actions, as I am also unable to reproduce this locally. 

Maybe let's wait to see if is fixed in some hours, and possibly file an issue to GH. :man_shrugging:

bin/fix_permissions
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -e
+# find all files where group or user are not odoo that are not a symlink, and fix permissions for those
+# we use find and xargs because chown because chown -R can be very slow on certain systems: https://github.com/docker/for-linux/issues/388
```
Yajo, line 4:
```suggestion
# we use find and xargs because chown -R can be very slow on certain systems: https://github.com/docker/for-linux/issues/388
```

bin/fix_permissions
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
Yajo, line 1: Would you mind converting this to python (at least, the same but with plumbum)? We have a goal here of getting rid of as much bash as possible (see #7).

That way we get more readability (the code is a bit cumbersome.. not that it could be better, bash is this way...), and code formatting, and all that stuff for free.

tests/__init__.py
```diff
@@ -457,7 +457,7 @@ def test_modified_uids(self):
                 (
                     "bash",
                     "-xc",
-                    'test "$(stat -c \'%U:%G\' /opt/odoo/custom/src)" == "root:odoo"',
+                    'test "$(stat -c \'%U:%G\' /opt/odoo/custom/src)" == "odoo:odoo"',
```
Yajo, line 460: This is a safety belt against malicious code: odoo and addons code is unmodifiable by the odoo user itself.

Sorry but this must stay as it was... 🙄

tests/__init__.py
```diff
@@ -486,7 +486,7 @@ def test_uids_mac_os(self):
                 (
                     "bash",
                     "-c",
-                    'test "$(stat -c \'%U:%g\' /opt/odoo/custom/src)" == "root:20"',
+                    'test "$(stat -c \'%U:%g\' /opt/odoo/custom/src)" == "odoo:20"',
```
Yajo, line 489: Also undo.

tests/__init__.py
```diff
@@ -514,7 +514,7 @@ def test_default_uids(self):
                 (
                     "bash",
                     "-xc",
-                    'test "$(stat -c \'%U:%G\' /opt/odoo/custom/src)" == "root:odoo"',
+                    'test "$(stat -c \'%U:%G\' /opt/odoo/custom/src)" == "odoo:odoo"',
```
Yajo, line 517: Also undo.

bin/autoaggregate
```diff
@@ -53,27 +93,13 @@ def aggregate(config):
...
             stderr=sys.stderr,
             stdout=sys.stdout,
+            preexec_fn=pre_exec,
         )
     finally:
-        # Restore umask, if changed
-        if old_umask is not None:
-            os.umask(old_umask)
         # Chown recursively, if UID or GID are specified
+        # just to make sure everything is owned by odoo:odoo (should run empty)
```
Yajo, line 100: This comment is probably fake (see below).

Yajo (changes requested):
The workaround seems good to me, although conceptually there's one change that we can't merge. It's commented below.

tests/__init__.py
ap-wtioit: ok, will change that, i was not sure because /opt/odoo/custom/src was chowned to odoo:odoo in `autoaggregate` and then chowned to root:odoo in 800-permissions

how do root:odoo permissions work with dev mode (/opt/odoo/custom mapped to the users working directory)?

bin/fix_permissions
ap-wtioit: I'm not sure plumbum would make this more readable but i can try.

bin/autoaggregate
ap-wtioit: What do you mean by fake comment (so i can improve it)?

bin/fix_permissions
joao-p-marques: One of the advantages here is that it could drop the pipelined commands and the usage of `xargs` and instead use Python to read the output from one command, parse it, and pass it to the other, making it a lot more readable.

ap-wtioit: well the point of this merge request is to make the build of scaffolding much faster (and therefore the tests). i'm pretty sure inventing our own xargs and find in python will not run as fast as just using the tools other people already worked hard on optimizing. in my experience custom code never runs as fast as the core linux utils. 

I'll still try to implement this. But i also then need to verify the speed again (and i already worked a few weeks for this on our systems), so please be patient with me.

joao-p-marques: Well, of course, if the lost in performance is big, it might not be worth it, but the idea @Yajo was mentioning was to use Plumbum to call those Linux tools (`find`, `chown`) and just use Python to control the flow and parse the results, that shouldn’t introduce a big overhead... 

> I'll still try to implement this. But i also then need to verify the speed again (and i already worked a few weeks for this on our systems), so please be patient with me.

Don’t worry, the contribution is very appreciated 🙂

tests/__init__.py
Yajo: > how do root:odoo permissions work with dev mode (/opt/odoo/custom mapped to the users working directory)?

The chown to `root:odoo` is done here: https://github.com/Tecnativa/doodba/blob/1c8d05aab51dfd21ed8a55af7ea30302cff614a9/build.d/800-permissions#L3

That's at build time.

However, in devel, your code is bind-mounted into the container, so whatever was in that directory at build time is overriden, and you can write to it.

Besides, [`invoke` tasks in the template automatically set UID, GID and UMASK](https://github.com/Tecnativa/doodba-copier-template/blob/72b97c1bafa69b9474ff948f08914e12d3a326f3/tasks_downstream.py#L21) to make `autoaggregate` behave as you would, outside the container, to avoid developer friction

bin/fix_permissions
Yajo: > i'm pretty sure inventing our own xargs and find in python will not run as fast as just using the tools other people already worked hard on optimizing

You can still do the pipelining in plumbum and use xargs too.

At the end of the day, `bash` spawns a new subprocess for every command. Replacing that with `python` shouldn't impact negatively in performance (at least I never noticed anything significant). However, an easier to read syntax makes code easier to reason about. And we have linters, formatters, and (more important) more human experience on that language, so it helps in the long run.

tests/__init__.py
ap-wtioit: Ah yes, that's right, somehow i didn't realize the difference between build and runtime mappings. Thanks for the explanation.

bin/autoaggregate
Yajo: Yes, sorry for the bare explanation.

I mean that not everything should be owned by `odoo:odoo`, but by `$UID:$GID`, in case they're specified. That's what the previous comment line says, and probably you were trying this chown because of https://github.com/Tecnativa/doodba/pull/413#discussion_r612978447.

So probably this comment is wrong.

bin/fix_permissions
ap-wtioit: Well as plumbum is python i wonder if some magic env variables for python have an effect on it. (E.g. -u or PYTHONUNBUFFERED to get stdout and stderr synced in python also seems to affect file io and i would expect also affecting pipelining in plumbum, but that's all on the test plan)

Yajo: Just one more detail. I just realized we have the `build.d/800-permissions` file which was meant to fix permissions.

Probably it would make more sense to move this code there.

bin/autoaggregate
```diff
@@ -53,27 +93,13 @@ def aggregate(config):
...
-                            # if target_path is a link we would only chown the
-                            # target, as the target should be chowned separately
-                            # we can skip all symlinks and prevent showing
-                            # unclear errors
-                            os.chown(target_path, UID, GID)
-                    except Exception:
-                        logger.debug(
-                            "Error trying to chown on file. Skipping...", exc_info=True
-                        )
+            check_call(["fix_permissions"])
```
Yajo, line 102: AFAICS, you're changing the permissions in the system beforehand to avoid chowning later.

If that's the case, what's the point on doing this call? 🤔

bin/autoaggregate
ap-wtioit: I was trying this chown because that's how it worked before (as long as i didn't missinterprete the code before):
* checkout everything as root:root in autoaggregate (by using root for gitaggregate), writing the data once
* chowning to odoo:odoo (`$UID:$GID`) with python in autoaggregate, writing the data twice
* later chowning to root:odoo with 800-permissions, writing the data a third time
And ideally with the fix/workarround the fix_permissions would not be needed anymore in aggregate (when we set the right permissions from the beginning). But unfortunately it still does change the .git (pack if i remeber correctly) readonly files from `r--r-----` to `rw-r-----`. Getting rid of this would make the code for fix_permissions even more complicated and unreadable and the performance impact was less than a second on our slow systems so it seemed not worth it.

So i would (if you agree that this is the right way)
* set the UID to 0 here for gitaggregate (to make everything owned by user root, and using git config from root)
* keep the GID code (to set the correct group right away, so everything should be assigned to group odoo)
* adapt  the comments (explaining that ideally fix_permissions should do nothing at this point)
* also i would try to make the doodba DEBUG log level switch own the -c output for chmod and -v output for chown

bin/autoaggregate
ap-wtioit: I am trying to make sure we have the right permissions after autoaggregate to prevent changing it later again. I think i could get rid of it but i was not 100% sure, so i kept fixing the permission here. (currently it only changes some permissions in .git from `r--r-----` to `rw-r-----`)

bin/fix_permissions
ap-wtioit: well i was moving it to a seperate bin file because i wanted consistent permissions across all steps (to prevent changing permissions to often) and as bin is present all the way (build + run time) in the container but build.d sounded like someday it could be present just during build time. so i didn't want bin/autoaggregate invoking something in build.d and instead invoke bin/fix_permissions from bin/autoaggregate and build.d/800-fix_permissions

ap-wtioit: Also it might just be a java habbit in me to not cross invoke package
s (build.d -> bin -> build.d -> bin ...) but instead have a clean one directional dependency (build.d -> bin but never bin -> build.d)

If that's not an issue for you i can jump over my own shadow and just invoke build.d/800-fix_permissions in bin/autoaggregate. (and it might not be needed after all)

bin/autoaggregate
ap-wtioit: Ok had some time to think the plan i mentioned above does not work:
* autoaggregate is also invoked with setup-devel.yaml to checkout locally, and permissions there need to match UID:GID of the invoking user, otherwhise the user is not the owner of his own files

How would you feel about another user (having a checkout user that can be the same UID as the developer but a different UID from the runtime odoo user (sharing the odoo group together))?

Yajo:
I'll answer to https://github.com/Tecnativa/doodba/pull/413#discussion_r612993147 here because I think it can solve more questions than just that one, and can be helpful as a reference on _why_ things are as they are. Maybe this will light some 💡 around 😄 

As you know, we support devel, test and prod environments.

> * checkout everything as root:root in autoaggregate (by using root for gitaggregate), writing the data once
> 
> * chowning to odoo:odoo (`$UID:$GID`) with python in autoaggregate, writing the data twice



The needs for devel are a bit specific: the volume must be bind-mounted in the container to develop comfortably. To make that possible, the developer needs RW access to all the files. And to make development faster, we don't want to force him to rebuild the image every time.

Thus, the `invoke git-aggregate` task calls the `autoaggregate` binary inside the container, with the specific UID, GID and UMASK required for it to work. Another gory detail is that [`invoke img-build` builds the image assigning to the inner `odoo` user the same UID and GID as yourself outside of the container](https://github.com/Tecnativa/doodba-copier-template/blob/72b97c1bafa69b9474ff948f08914e12d3a326f3/devel.yaml.jinja#L35-L37). Hopefully some day we'll be able to move development to podman and forget about all these things, but with docker it's what we have.

> later chowning to root:odoo with 800-permissions, writing the data a third time

The important bit here is that `800-permissions` runs on builds, while `autoaggregate` can run when a container was already built (in devel only, of course).

When we're building for test or prod, changing odoo user UID/GID is a mess because it would conflict with the UID/GID inside the `backup` container (for example), and possibly other things, so it's always 1000 (although it can be changed, but it's not really recommended).

As such, when we're building test or prod, those env variables don't really matter for `autoaggregate`, because anyways we're gonna chown it later in the same build (on `800-permissions`).

Finally (this was already told, but it completes the picture), the fact of using `root:odoo` is a safety belt that effectively forbids a malicious addon to modify odoo code.

----

So, after putting all of these words together, I think that we can safely get rid of `fix_permissions` or `800-permissions`, because:

1. For devel, we can just modify UID/GID/UMASK before doing the operations, and that will simply work, without needing to change anything. That is container runtime.
2. For test/prod, we can do the same as for devel, but exporting `UID=root GID=odoo UMASK=u=rwX,g=rX,o=` inside `100-repos-aggregate` to make sure we aggregate properly in the 1st place. The latter chown in `800-permissions` would do nothing, so it could be removed. That is image build time.

Or maybe you have a better idea?

ap-wtioit:
That sound's good, (i already tested a variant of exporting UID=0 inside 100-repos-aggregate before autoaggregate, but the full export including umask sound's much better) i will test that and update the pull request when it works (not if ;-) )

ap-wtioit:
TODOs @ap-wtioit:
- [ ] add a test / ensure there is a test for devel autoaggregate permissions (autoaggregate run as odoo:odoo during runtime)
- [ ] remove fix_permissions calls where not needed
- [ ] move fix_permissions code back into 800-fix_permissions
- [ ] rework 800-fix_permissions to plumbum

ap-wtioit:
Template for change announcement in https://github.com/Tecnativa/doodba/discussions/67:
```Markdown
# Change of UMASK integer format
## When will this happen?

Once #413 is merged, UMASK needs to be given as an octal integer.


## Why will this happen?

Umasks in linux usually are given as an octal integer when in integer form. Also all documentation mentioned a UMASK that pointed to octal integer form. But during autoaggregate the string given as UMASK was read a decimal integer which would give unexpected results.

## What does it break?

Users specifying a UMASK as decimal integer may get the wrong permissions after this change. Users following the documentation and specifying the umask as octal integer (e.g. `UMASK=$(umask)`, `UMASK=27` or using https://github.com/Tecnativa/doodba-copier-template/blob/main/tasks_downstream.py) should get the correct permissions with this change.


Previously linux umask 0027 (permit nothing for others, do not permit write for group) needed to be given as `UMASK=23`. Whereas `UMASK=27` would have been linux umask 0033 (do not permit read and write for group and other (but allow execute for group and others))

## Last non-breaking image/template version

TBD
```

ap-wtioit:
@Yajo @joao-p-marques sorry it took so long. from my point of view this would now be ready.

joao-p-marques:
Hi @ap-wtioit :wave: 

Thanks again for the work and contribution :+1: 
Let's just wait a few days until @Yajo gets a chance to review again. I'll also try to review and test in the next days.

13.0.Dockerfile
```diff
@@ -210,19 +222,8 @@ ONBUILD ENV ADMIN_PASSWORD="$ADMIN_PASSWORD" \
...
-
-# Enable Odoo user and filestore
-ONBUILD RUN groupadd -g $GID odoo -o \
-    && useradd -l -md /home/odoo -s /bin/false -u $UID -g $GID odoo \
-    && mkdir -p /var/lib/odoo \
-    && chown -R odoo:odoo /var/lib/odoo /qa/artifacts \
-    && chmod a=rwX /qa/artifacts \
-    && sync
+ONBUILD COPY --chown=root:odoo $LOCAL_CUSTOM_DIR /opt/odoo/custom
+ONBUILD RUN /opt/odoo/common/build.d/800-permissions
```
Yajo, line 226: If I'm not mistaken, you don't need this line, because it will get executed later when running `/opt/odoo/common/build`, right?

Also, I don't see it in other dockerfiles.

Yajo (changes requested):
All seems pretty good, thanks!

I only saw one thing that I think must have been forgotten in the dev loop, please check below.

@joao-p-marques please review.

13.0.Dockerfile
ap-wtioit: yes, you are right, should be removed here as well.

Yajo (approved):

joao-p-marques (commented):
Thanks again for the contribution @ap-wtioit :+1: 

Just some minor comments (mostly doubts :smile:). Otherwise, tested locally and everything looks good, it is taking nearly 1 minute less to build in my tests for v14 :+1: :tada:

bin/autoaggregate
```diff
@@ -33,47 +34,61 @@ def aggregate(config):
...
+            "--log-level",
+            log_level,
+            "--jobs",
+            str(cpu_count() or 1),
+            "aggregate",
+        ],
+        cwd=SRC_DIR,
+        stderr=sys.stderr,
+        stdout=sys.stdout,
+        preexec_fn=pre_exec,
```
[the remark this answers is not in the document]
ap-wtioit: I would not think so, `your application` in this call should be `autoaggregate` which doesn't use threads. From `autoaggregate`s perspective it doesn't matter that gitaggregate uses multiple threads as we are launching it as a single process/command.

bin/autoaggregate
```diff
@@ -33,47 +34,61 @@ def aggregate(config):
...
+    def pre_exec():
+        for _exec in pre_execs:
+            try:
+                _exec()
+            except Exception as e:
+                logger.error("Error in %s: %s" % (_exec, e))
+                logger.exception(e)
+                raise
+
+    if UMASK:
```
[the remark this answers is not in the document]
ap-wtioit: `UMASK` can not be evaluated `False` but cloud be falsy. E.g. you can launch autoaggregate with `UMASK="" autoaggregate` to prevent setting a umask for checkout.

bin/autoaggregate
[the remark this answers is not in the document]
joao-p-marques: I see, thanks :+1:

bin/autoaggregate
```diff
@@ -33,47 +34,58 @@ def aggregate(config):
         Path where to find the ``repos.yaml`` file.
     """
     logger.info("Running gitaggregate with %s", config)
-    old_umask = None
-    try:
+
+    def pre_exec_umask():
         # Download git code with the specified umask, if any
```
joao-p-marques, line 39:
```suggestion
        # Download git code with the specified umask, if set, otherwise use "0027"
```

joao-p-marques (approved):
Thanks :+1:

bin/autoaggregate
ap-wtioit: thanks, yeah that's better

wtaferner (approved):
😄

PR #444, opened 2021-09-09
Prepare v15

joao-p-marques:
Prepares base for v15, also making the necessary changes to make it work with the current `saas-14.5`. Once the branch `15.0` is available in `OCA/OCB`, we can just revert 433dbb3

@Tecnativa
TT17396

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
-.vscode
 **.pyc
--- a/15.0.Dockerfile
+++ b/15.0.Dockerfile
@@ -1,0 +1,233 @@
+FROM python:3.9-slim-buster AS base
+
+EXPOSE 8069 8072
+
+ARG GEOIP_UPDATER_VERSION=4.3.0
+ARG WKHTMLTOPDF_VERSION=0.12.5
+ARG WKHTMLTOPDF_CHECKSUM='1140b0ab02aa6e17346af2f14ed0de807376de475ba90e1db3975f112fbd20bb'
+ENV DB_FILTER=.* \
+    DEPTH_DEFAULT=1 \
+    DEPTH_MERGE=100 \
+    EMAIL=https://hub.docker.com/r/tecnativa/odoo \
+    GEOIP_ACCOUNT_ID="" \
+    GEOIP_LICENSE_KEY="" \
+    GIT_AUTHOR_NAME=docker-odoo \
+    INITIAL_LANG="" \
+    LC_ALL=C.UTF-8 \
+    LIST_DB=false \
+    NODE_PATH=/usr/local/lib/node_modules:/usr/lib/node_modules \
+    OPENERP_SERVER=/opt/odoo/auto/odoo.conf \
+    PATH="/home/odoo/.local/bin:$PATH" \
+    PIP_NO_CACHE_DIR=0 \
+    DEBUGPY_ARGS="--listen 0.0.0.0:6899 --wait-for-client" \
+    DEBUGPY_ENABLE=0 \
+    PUDB_RDB_HOST=0.0.0.0 \
+    PUDB_RDB_PORT=6899 \
+    PYTHONOPTIMIZE=1 \
+    UNACCENT=true \
+    WAIT_DB=true \
+    WDB_NO_BROWSER_AUTO_OPEN=True \
+    WDB_SOCKET_SERVER=wdb \
+    WDB_WEB_PORT=1984 \
+    WDB_WEB_SERVER=localhost
+
+# Other requirements and recommendations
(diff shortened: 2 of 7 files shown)


pedrobaeza:
Wait until knowing Python version from Odoo.

joao-p-marques:
Ok, we'll leave it as draft in the meantime :+1:

.gitignore
```diff
@@ -1,2 +1 @@
-.vscode
```
Yajo, line 1: Undo this please. The settings file is forced, so it works fine as it was. But vscode stores some garbage there that we don't want to track by default.

Alternatively you can use this template: https://www.toptal.com/developers/gitignore/api/visualstudiocode

15.0.Dockerfile
```diff
@@ -93,8 +93,8 @@ RUN python -m venv --system-site-packages /qa/venv \
     && deactivate \
     && mkdir -p /qa/artifacts
 
-ARG ODOO_SOURCE=OCA/OCB
-ARG ODOO_VERSION=15.0
+ARG ODOO_SOURCE=odoo/odoo
+ARG ODOO_VERSION=saas-14.5
```
Yajo: Add a TODO.

15.0.Dockerfile
```diff
@@ -176,7 +176,7 @@ ONBUILD ENTRYPOINT ["/opt/odoo/common/entrypoint"]
 ONBUILD CMD ["/usr/local/bin/odoo"]
 ONBUILD ARG AGGREGATE=true
 ONBUILD ARG DEFAULT_REPO_PATTERN="https://github.com/OCA/{}.git"
-ONBUILD ARG DEFAULT_REPO_PATTERN_ODOO="https://github.com/OCA/OCB.git"
+ONBUILD ARG DEFAULT_REPO_PATTERN_ODOO="https://github.com/odoo/odoo.git"
```
Yajo: Add a TODO.

hooks/build
```diff
@@ -20,8 +20,6 @@ for target, suffix in (("base", ""), ("onbuild", "-onbuild")):
         "VCS_REF=%s" % COMMIT,
         "--build-arg",
         "BUILD_DATE=%s" % date().strip(),
-        "--build-arg",
-        "ODOO_VERSION=%s" % VERSION,
```
Yajo, line 24: Add a TODO.

Yajo (changes requested):
Looks good overall.

Please add some TODO comments instead (or apart) of relying on a commit, to make this pending state more obvious when reading the code.

Besides that, could you confirm this please?

- Have you checked `ImportError` catches across v15 codebase to see if there are new soft dependencies?
- Have you checked whether the soft dependencies we were installing before still are necessary?

joao-p-marques:
> The diff is a bit confusing, IMHO don't remove v11 yet, and leave it there until it starts giving problems.

Ok, I restored it.

About the diff, I will leave here a diff between the Dockerfile from version 14 and 15:

```diff
1c1
< FROM python:3.8-slim-buster AS base
---
> FROM python:3.8-slim-bullseye AS base
49d48
<         npm \
53d51
<         zlibc \
87,90d84
<         flake8 \
<         pylint-odoo \
<         six \
<     && npm install --loglevel error --prefix /qa eslint \
95c89
< ARG ODOO_VERSION=14.0
---
> ARG ODOO_VERSION=15.0
```

Please notice that, following @ap-wtioit suggestions, I did some updates to the 13 and 14 Dockerfiles to install the appropriate packages for the current Debian release.

ap-wtioit:
For debian bullseye ruby version (`ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-linux-gnu]`) `gem` no longer supports `--no-rdoc --no-ri` i guess we need to switch to `--no-document` in [lib/doodbalib/installler.py](https://github.com/Tecnativa/doodba/blob/b2286b8733c876ec0f97bd01ee9cc346af6648fb/lib/doodbalib/installer.py#L91)

All ruby versions we currently use
* 15.0 bullseye: `ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-linux-gnu]`
* 13.0, 14.0 buster: `ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-linux-gnu]`
* 11.0, 12.0 stretch: `ruby 2.3.3p222 (2016-11-21) [x86_64-linux-gnu]`
support the `--no-document` option instead of `--no-rdoc --no-ri`

ap-wtioit:
When removing Node.js (npm) from images the `test-npm-install` stuff also needs to be disabled for those versions (13.0, 14.0, 15.0)

PR #453, opened 2021-10-18
Use git-autoshare with git-aggregator

joao-p-marques:
Take advantage of https://github.com/acsone/git-autoshare to make the git aggregation process faster and more efficient with disk space.

Related to:

- [ ] https://github.com/acsone/git-aggregator/pull/55
- [ ] https://github.com/acsone/git-autoshare/pull/15

@Tecnativa
TT32042

### DIFF
--- a/13.0.Dockerfile
+++ b/13.0.Dockerfile
@@ -13,6 +13,10 @@ ENV DB_FILTER=.* \
     GEOIP_ACCOUNT_ID="" \
     GEOIP_LICENSE_KEY="" \
     GIT_AUTHOR_NAME=docker-odoo \
+    GIT_AUTHOSHARE=0 \
+    GIT_AUTOSHARE_CACHE_DIR="/home/odoo/.cache/git-autoshare/" \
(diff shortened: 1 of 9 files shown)


yajo (changes requested):
CI shows errors:

    git-autoshare  /home/odoo/.config/git-autoshare/repos.yml  not found. No hosts to load.

That seems like a bug in git-autoshare. I think it should just skip itself if there's no repos definition, isn't it?

Also please add docs about how to benefit from this for example using CI caching.

And a test also.

OTOH great addition!

yajo (changes requested):

pedrobaeza:
@david-banon-tecnativa can you resume this one to check if this is still improving something and the caveats it has?

david-banon-tecnativa:
I started working on this during the OCA days actually, i don't think we can do it to speed up CI, but we should be able to see a improvement in local build times. I don't know for sure if it will help with storage size.

yajo:
You might want to consider replacing git-autoshare with [mrchef](https://gitlab.com/moduon/mrchef), which uses git-autoshare out of the box if available, and supports lockfiles.

Issue #389, opened 2021-02-13
[Feature Request] Independent wkhtmltopdf image


amon-ra:
This [project](https://github.com/acsone/kwkhtmltopdf) implement a client-server architecture to wkhtmltopdf. It reduces the dependencies of the base image and it makes easy to distribute load.

¿Do you think it could be used?

Yajo:
It seems nice, although #342 is not on our roadmap in any case.

amon-ra:
I will try to send a pull request

amon-ra:
Done. Hope it helps.

PR #494, opened 2022-04-07
Independent wkhtmltopdf

amon-ra:
This pull try to solve [Tecnativa/doodba#4039](https://github.com/Tecnativa/doodba/issues/389) . It will load wkhtmltopdf from an external docker image in port 8080.


theangryangel:
Forgive me if I’m wrong, as I’m reading this PR on my phone, but this looks like a major breaking change?

As a very grateful user of Doodba, I like the idea of this in theory, but if it could be opt in and keep the existing functionality I feel it would be much more likely to be merged…

amon-ra:
Now you can use build arg WKHTMLTOPDF_SKIP=1 to skip wkhtmltopdf build and use the new binary.

bosd:
@amon-ra  Can you re-trigger tests?

PR #497, opened 2022-04-26
feat: update db expiration date when preparing db

yajo:
This script is called when using `invoke preparedb` in develop mode.

If developing enterprise code, this avoids an unnecessary warning.


Shide (approved):

Tardo:
Would it be possible to check that this https://github.com/odoo/enterprise/blob/bbc68dd2e73615711fe325c57dfaaaa4fbf0cca7/web_enterprise/version.py#L9 is being applied before writing this value? This way we only write in enterprise environments.

yajo:
I don't think it's necessary because this shouldn't harm other environments and the invoke tasks only get executed on development.

PR #508, opened 2022-07-04
[FIX] direxec should not try to execute directories

ap-wtioit:
In f931b534872fed96e9037ad31eb5b90ada4b3111 direxec was converted from bash to python. In the python version we should not try to execute directories. As the execute flag for directories only means that the user can change into the directory not that the directory is an executable.

This happens when a user adds a custom python script to entrypoint.d and runs doodba built with `COMPILE: "true"`. for this case a `__pycache__` directory is created in `/opt/odoo/custom/entrypoint.d/` that causes the following error:
```
PermissionError: [Errno 13] Permission denied: '/opt/odoo/custom/entrypoint.d/__pycache__'
```

Info @wt-io-it

### DIFF
--- a/bin/direxec
+++ b/bin/direxec
@@ -25,7 +25,7 @@ except OSError:
 # Run scripts
 for executable, folder in sorted(files):
     command = os.path.join(folder, executable)
-    if os.access(command, os.X_OK):
+    if os.access(command, os.X_OK) and not os.path.isdir(command):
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Can you please rebase?

ap-wtioit:
rebased and all checks still passing

pedrobaeza (approved):

PR #512, opened 2022-07-14
[IMP] support customization of the custom conf path.

LoisRForgeFlow:
This allow you to potentially build different configuration files
based on the enviroment you are using.

@ForgeFlow


ap-wtioit (commented):
A test would be nice (test_dotd has a fragment that tests custom/conf.d maybe we could expand this)
```
                # ``custom/conf.d`` was properly concatenated
                ("grep", "test-conf", "auto/odoo.conf"),
```

`./custom` is copied onbuild into `/opt/odoo/custom` and is chowned to `root:odoo`. In dev adding additional directories should be fine (if one exports UID, GID properly). but in other environments there could be permission issues with reading custom conf.d that do not belong to odoo uid inside the container.

Can we ignore such permission issues?

PR #523, opened 2022-09-21
[FIX] git-aggregator: use git-aggregator < 3.0.0 for V11.0 and V12.0

ap-wtioit:
filter=blob:none is not compatible with older git version in debian stretch

After git-aggregator 3.0.0 got relased our nightly CI builds failed for 11.0 and 12.0 with:
```
(ERROR) [21:20:47] git_aggregator.repo  misc-addons   /opt/odoo/custom/src/misc-addons> error calling ('git', 'clone', '--filter=blob:none', '--depth', '1', 'https://github.com/itpp-labs/misc-addons.git', '/opt/odoo/custom/src/misc-addons')
error: unknown option `filter=blob:none'
```

Info @wt-io-it

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -82,7 +82,7 @@ RUN pip install \
         astor \
         # Install fix from https://github.com/acsone/click-odoo-contrib/pull/93
         git+https://github.com/Tecnativa/click-odoo-contrib.git@fix-active-modules-hashing \
-        git-aggregator \
+        "git-aggregator<3.0.0" \
         "pg_activity<2.0.0" \
         plumbum \
         ptvsd \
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Not sure why the CI doesn't finish...

ap-wtioit:
> Not sure why the CI doesn't finish...

those builds are no longer / not defined in the build matrix:
https://github.com/Tecnativa/doodba/blob/master/.github/workflows/ci.yaml#L34
so they will not be started. i'm not sure where github picks up that they are required

all the jobs that are defined in ci.yaml already finished

pedrobaeza:
I have pushed 7b25d25d8e56beee2592b744facf9e518ffca2e6 with your same patch, but for all versions, while https://github.com/acsone/git-aggregator/issues/68 is not solved.

PR #528, opened 2022-11-08
[FIX] stretch: use apt-archive for postgres in stretch

ap-wtioit:
https://www.postgresql.org/message-id/Y2kmqL%2BpCuSZiQBV%40msg.df7cb.de

Postgres no longer maintains debian stretch on their apt repos and moved the old packages to apt-archive.

On our CI this failed with the following error before the fix:
```
W: The repository 'http://apt.postgresql.org/pub/repos/apt stretch-pgdg Release' does not have a Release file.
E: Failed to fetch http://apt.postgresql.org/pub/repos/apt/dists/stretch-pgdg/main/binary-amd64/Packages  404  Not Found [IP: 147.75.85.69 80]
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

info @wt-io-it

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -50,7 +50,9 @@ RUN apt-get -qq update \
         telnet \
         vim \
         zlibc \
-    && echo 'deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main' >> /etc/apt/sources.list.d/postgresql.list \
+        apt-transport-https \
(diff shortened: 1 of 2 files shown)


ap-wtioit:
updated urls to http (instead of https) because of:
```
#6 78.48 E: The method driver /usr/lib/apt/methods/https could not be found.
```

pedrobaeza (approved):

PR #560, opened 2023-08-24
[IMP] use PY 3.9 for avoid issue at install poetry

celm1990:
actions is not working because running over Ubuntu-latest with Python 3.11
I tested on Py 3.10 and same error, but into PY 3.9 is working
image


pedrobaeza (approved):
Change minimum PG version for 13.0 and 12.0 and working now working.

PR #568, opened 2023-11-05
Add Odoo 17.0

pedrobaeza:
- Dockerfile copied from 16.0 and slightly adapted
- Fix tests
- Make it `:latest`.
- Deprecate some old Odoo versions.
- Bump PG to 15.

TODO: Switch to Python 3.10

@Tecnativa TT45808

### DIFF
--- a/17.0.Dockerfile
+++ b/17.0.Dockerfile
@@ -1,0 +1,232 @@
+FROM python:3.10-slim-bullseye AS base
+
+EXPOSE 8069 8072
+
+ARG GEOIP_UPDATER_VERSION=4.3.0
+ARG WKHTMLTOPDF_VERSION=0.12.5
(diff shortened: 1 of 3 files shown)


17.0.Dockerfile
```diff
@@ -0,0 +1,232 @@
+FROM python:3.8-slim-bullseye AS base
```
celm1990, line 1: Odoo 17 need minimum  PY 3.0 according this commit
https://github.com/odoo/odoo/commit/163fb3ae618bf993e60f8c1276e09a5b85565fa5
Do you sure this working with 3.8?

pedrobaeza: I see, we will need to make it work with P3.10. I was delaying such step, as it was conflicting for 16 (and still not working).

celm1990: I tested into commit https://github.com/celm1990/doodba/commit/888f30b6fca66b54daba167a6f3917d89e68f324 and it's working now

pedrobaeza:
Reopening other PR for avoiding obsolete branches marked as required.

PR #575, opened 2023-11-10
[FIX] Fix tests for 17.0

josep-tecnativa:
There are problems in next tests:
- test_dependencies_base_search_fuzzy
- test_dotb
- test_addons_filtered 

CC @Tecnativa TT45808


tests/scaffoldings/dotd/custom/dependencies/pip.txt
```diff
@@ -1,6 +1,5 @@
 # Compiled locally, which would fail if `apt_build.txt` fails
+pip==23.3.1
```
ap-wtioit, line 2: the comment above this line belongs to the pycrypto pip line, if you are locking the pip version please add it as line 1 (and maybe add a comment why the version is locked)

initially each of the lines in this pip.txt worked as a separate test on their own here (as far as i remember)

josep-tecnativa: Certainly. I added this line (version 23.3.1, it's the latest version of pip) because I wanted to make sure that the pip version wouldn't cause any issues. I am testing locally to confirm if I don't need it, and if that's the case, I'll remove it. Thanks for the review!

pedrobaeza: We are using for now an old Debian version. Another thing to do is to use bookworm instead.

.vscode/settings.json
```diff
@@ -12,5 +12,6 @@
   "[dockerfile]": {
     "editor.formatOnSave": false
   },
-  "python.analysis.extraPaths": ["./lib", "./tests"]
+  "python.analysis.extraPaths": ["./lib", "./tests"],
+  "python.languageServer": "None"
```
pedrobaeza, line 16: Why changing this?

josep-tecnativa: Sorry, it was done automatically by my vscode. I think that this PR is ready to merge, as the tests have passed before this last change.

PR #578, opened 2023-11-15
[IMP] Update debian to bookworm in v17.0 and v16.0 

josep-tecnativa:
CC @Tecnativa TT45983


pedrobaeza:
Rebase for solving conflicts, and better to start just with one version for minimizing problems. Always one problem at a time.

17.0.Dockerfile
```diff
@@ -36,7 +36,7 @@ ENV DB_FILTER=.* \
 RUN apt-get -qq update \
     && apt-get install -yqq --no-install-recommends \
         curl \
-    && curl -SLo wkhtmltox.deb https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/${WKHTMLTOPDF_VERSION}/wkhtmltox_${WKHTMLTOPDF_VERSION}-1.buster_amd64.deb \
+    && curl -SLo https://github.com/wkhtmltopdf/packaging/releases/download/${WKHTMLTOPDF_VERSION}-3/wkhtmltox_${WKHTMLTOPDF_VERSION}-3.bookworm_amd64.deb \
```
ap-wtioit, line 39: did you test if the Odoo S.A. maintained version of wkhtmltopdf works here?
https://github.com/odoo/wkhtmltopdf

Would be nice if we can switch to the one built for Odoo.

josep-tecnativa: Oh, okay thank you! I am going to change to the odoo wkhtmltopdf nightly build

pedrobaeza: Other option is to use dockerized wkhtmltopdf: https://github.com/acsone/kwkhtmltopdf. That way, we also isolate possible security issues.

josep-tecnativa:
Please @pedrobaeza , review this changes. I will open other PR in order to dockerize wkhtmltopdf if necessary

17.0.Dockerfile
```diff
@@ -230,4 +230,3 @@ ONBUILD ARG DB_VERSION=latest
 ONBUILD RUN /opt/odoo/common/build && sync
 ONBUILD VOLUME ["/var/lib/odoo"]
 ONBUILD USER odoo
-CMD ["sh", "-c", "ls || sleep infinity"]
```
pedrobaeza: Please remove this command on the commit where it was added, not polluting the real diff.

tests/__init__.py
```diff
@@ -406,10 +513,97 @@ def test_dotd(self):
                 ("--version",),
             )
 
+    def test_dotd_ge_16(self):
+        """Test environment with common ``*.d`` directories.
+        Only for versions greater or equal to 16"""
+        for sub_env in matrix(odoo_skip={"11.0", "12.0", "13.0", "14.0", "15.0"}):
```
pedrobaeza, line 519: Please avoid duplicating all these stuff. You can do inside the same test several loops, letting the common ones together, and just including the differences in different loops.

pedrobaeza: Adding new commits amending code from this same PR makes the review harder. Better to cleanup the commits that add the original code.

PR #581, opened 2023-11-20
[FIX] flanker new version dependencies

pedrobaeza:
The newest flanker version includes 2 dependencies that are not automatically installed, so we need to explicitly add them.

@Tecnativa

### DIFF
--- a/16.0.Dockerfile
+++ b/16.0.Dockerfile
@@ -123,6 +123,8 @@ RUN build_deps=" \
         click-odoo-contrib \
         debugpy \
         pydevd-odoo \
+        redis \
+        dnsq \
         git+https://github.com/mailgun/flanker.git@v0.9.15#egg=flanker \
         geoip2 \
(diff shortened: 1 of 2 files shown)


16.0.Dockerfile
```diff
@@ -123,6 +123,8 @@ RUN build_deps=" \
         click-odoo-contrib \
         debugpy \
         pydevd-odoo \
+        redis \
+        dnsq \
         git+https://github.com/mailgun/flanker.git@v0.9.15#egg=flanker \
```
yajo, line 128: You could have done this instead:

```suggestion
        git+https://github.com/mailgun/flanker.git@v0.9.15#egg=flanker[validator] \
```

Thanks anyway for the quick fix!

pedrobaeza: #584 opened for switching to this better form (in case they change in the future).

PR #588, opened 2023-11-22
[FIX] 16/17: Proper ~root/.ssh pre-removal

mfloridosygel:
Hello, 

I found that the symlink to /opt/odoo/custom/ssh still being created as sub-directory (/root/.ssh/ssh) so I got an error when I tried to  build my Odoo image.

Docker uses "/bin/sh" as default shell and "[[ ]]" is a specific comparison construct of Bash.

I remove the brackets and now the symlinks to /opt/odoo/custom/ssh is being created succesfully.

Issue origin: https://github.com/Tecnativa/doodba/pull/585

@pedrobaeza

### DIFF
--- a/16.0.Dockerfile
+++ b/16.0.Dockerfile
@@ -230,7 +230,7 @@ ONBUILD COPY --chown=root:odoo $LOCAL_CUSTOM_DIR /opt/odoo/custom
 
 # https://docs.python.org/3/library/logging.html#levels
 ONBUILD ARG LOG_LEVEL=INFO
-ONBUILD RUN [[ -d ~root/.ssh ]] && rm -r ~root/.ssh; \
+ONBUILD RUN [ -d ~root/.ssh ] && rm -r ~root/.ssh; \
             mkdir -p /opt/odoo/custom/ssh \
             && ln -s /opt/odoo/custom/ssh ~root/.ssh \
             && chmod -R u=rwX,go= /opt/odoo/custom/ssh \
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Great, thanks for the fix.

Good point using a regexp for indicating the versions, but better to be explicit and as brackets are used for separators, it seems weird that way. Please rephrase it to `[FIX] 16/17: Proper ~root/.ssh pre-removal`, and you can put as extra commit message lines the explanation given in this PR.

mfloridosygel:
Done :ok_hand: 

Thank you!

pedrobaeza:
The explanation given in the first comment of the PR (which is pretty good) has not been put in the commit message. It's also better to use a neutral redaction avoiding first person. Something like:

```
[FIX] 16/17: Proper ~root/.ssh pre-removal

The symlink to /opt/odoo/custom/ssh is still being created as sub-directory 
(/root/.ssh/ssh), getting an error when trying to build the Doodba image.

Docker uses "/bin/sh" as default shell and "[[ ]]" is a specific comparison
construct of Bash.

Removing the brackets, now the symlinks to /opt/odoo/custom/ssh are being
successfully created.
```

mfloridosygel:
Hello,

Sorry I did not understand you at the first time.

Now it is done :+1: . 

Thank you!

@pedrobaeza

ap-wtioit (approved):
Good catch.

pedrobaeza (approved):

Issue #589, opened 2023-11-23
Excesive top margin downloading PDF documents


mfloridosygel:
## Describe the bug

When you download a PDF (e.g. Preview or Invoice) for the first time it appears correctly, but when you download it again, the top margin increases.

It happens using doodba-copier-template and my private projects.

Paper format is not modified, all by default.

The parameter "report.url" is correctly configurated.
![image](https://github.com/Tecnativa/doodba/assets/127136046/654fafcd-5e78-4913-aad2-d9f4a1580f47)

First download:
![image](https://github.com/Tecnativa/doodba/assets/127136046/606377c6-0f94-4fa9-aff5-37881be0b1f8)

Second download:
![image](https://github.com/Tecnativa/doodba/assets/127136046/5ca8e1f1-83d9-44b0-94f9-42966e39cc69)

## To Reproduce

I am using doodba-copier-template, choosing Odoo version "16.0" and Postgresql version '13'.

Here is my copier-answers,yml:
```
# Changes here will be overwritten by Copier; NEVER EDIT MANUALLY
_commit: v5.3.5
_src_path: gh:Tecnativa/doodba-copier-template
backup_dst: ''
cidr_whitelist: null
domains_prod: {}
domains_test: {}
gitlab_url: ''
odoo_dbfilter: ^prod
odoo_initial_lang: es_ES
odoo_listdb: true
odoo_oci_image: ''
odoo_proxy: traefik
odoo_version: 16.0
paths_without_crawlers:
- /web
- /website/info
postgres_dbname: prod
postgres_exposed: false
postgres_username: odoo
postgres_version: '13'
project_author: Tecnativa
project_license: BSL-1.0
project_name: myproject-odoo
smtp_default_from: ''
smtp_relay_host: '''
```

**Affected versions**:
It affects on version 16.0 and 17.0.

1. Command: "copier copy --trust gh:Tecnativa/doodba-copier-template project/folder".
2. Command: "invoke develop img-build git-aggregate resetdb start".
3. Activate "sale_management" module.
4. Go to "General settings" > "Companies" > "Document Layout" >  "Configure Document Layout".
5. Paper format > A4.
6. Click on "Download PDF Preview".

The first one will be correct, but the second will have an excesive top margin.

**Expected behavior**
All the downloads like the first one, with a normal top margin.

**Additional context**
Ubuntu 22.04.
Docker version 24.0.7, build afdd53b.

HaraldPanten:
[...]

PR #590, opened 2023-11-24
[FIX] Change version wkhtmltopdf 0.13 to 0.12.6.1 to avoid printig problems

josep-tecnativa:
CC @Tecnativa TT46155

This PR solves the issue https://github.com/Tecnativa/doodba/issues/589


HaraldPanten:
@josep-tecnativa might be useful to link the original issue in the commit message for more info. WDYT?

PR #599, opened 2024-02-28
[FIX] Temporal fix meanwhile odoo uses -intra option in ffmpeg. This …

josep-tecnativa:
…option is not compatible with ffmpeg lasts versions.
https://github.com/OCA/OCB/blob/52bec035f3afbe4e1cc9e51e062ed35eb830e04f/odoo/tests/common.py#L1360
@Tecnativa TT48083

### DIFF
--- a/16.0.Dockerfile
+++ b/16.0.Dockerfile
@@ -49,7 +49,6 @@ RUN echo "LAST_SYSTEM_UID=$LAST_SYSTEM_UID\nLAST_SYSTEM_GID=$LAST_SYSTEM_GID\nFI
     && apt-get install -yqq --no-install-recommends \
         ./wkhtmltox.deb \
         chromium \
-        ffmpeg \
         fonts-liberation2 \
         gettext \
         git \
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Put the reference to Odoo code using that option.

chienandalu (commented):
chore: Add some extra rationale to the commit message:

Debian bookworm and upper versions use version 5 of ffmpeg (used in video test tours artifacts generation) which doesn't support the `intra` flag used by Odoo (link).

This options has been deprecated for long time and we've proposed a fix (link to pr) for a modern alternative. While these changes aren't incorporated we'll force the version to support the current options.

chienandalu:
Thanks @ap-wtioit !

chienandalu:
Once that Odoo PR is merged maybe we should add that test to the doodba CI

josep-tecnativa:
> Once that Odoo PR is merged maybe we should add that test to the doodba CI

Meanwhile, I'm working in this PR as a temportal patch. Now I'am facing some compatibility issues between 4.x ffmepg version and bookworm, so we can't merge this PR yet.

ap-wtioit:
> Meanwhile, I'm working in this PR as a temportal patch. Now I'am facing some compatibility issues between 4.x ffmepg version and bookworm, so we can't merge this PR yet.

Another fix, would be to add a script `ffmpeg`  into the odoo users path that accepts the `subprocess.run([ffmpeg_path, '-intra', '-f', 'concat','-safe', '0', '-i', concat_script_path, '-pix_fmt', 'yuv420p', outfile])` call and converts it to `subprocess.run([ffmpeg_path, '-f', 'concat', '-safe', '0', '-i', concat_script_path, '-pix_fmt', 'yuv420p', '-g', '0', outfile], check=True)`.

E.g. (not tested):
```bash
#!/usr/bin/env bash
if [[ $1 == "-intra" ]] ; then
  # discard first argument -intra
  shift
fi
exec /usr/bin/ffmpeg "$@"
```
if you put this as ffmpeg into /home/odoo/.local/bin as ffmpeg it should remove the `-intra` parameter before executing the real ffmpeg. if we also need the added `-g 0` i can help you with the wrapper (On monday).

pedrobaeza:
Let's wait for the following days to have the Odoo PR merged, and if no response, then let's go to alternative solutions.

Issue #600, opened 2024-03-02
[17.0] Geoip not working correctly


shaneo1:
I have installed Odoo 17 using Doodba and use global price lists and have used this in odoo 16 with the geolite2 database. But was only able to get it working using a thirdparty app from odoo.com. 

I have installed odoo 17 using a different docker-compose script that puts both Geolite2-city and country in to /usr/share/GeoIP as instructed my the odoo documentation.

Now in Doodba there is on the city database and the odoo.conf in pointing to its location geoip_database = /opt/odoo/auto/geoip/GeoLite2-City.mmdb. and the geoip does not work for the country,  I believe this is a bug in the scripting as I am unable to detect the customer country against the IP address. How can I rectify this issue.

ap-wtioit:
@shaneo1 did you setup the geoip credentials correctly? (is there a database in /opt/odoo/auto/geoip/GeoLite2-City.mmdb?)

shaneo1:
There is a city database yes.

shaneo1:
> Ok just investigated (because we have an issue with the [new download limits for GeoIP](https://dev.maxmind.com/geoip/release-notes/2024#api-policies---temporary-enforcement-on-february-7-2024) and testing on our CI) and i can confirm the parameters for GeoIP in Odoo have changed in Odoo 17.0 and are no longer compatible with the doodba implementation:
> 
> [Odoo 17.0](https://github.com/odoo/odoo/blob/17.0/odoo/tools/config.py#L318)
> 
> ```
>         group.add_option("--geoip-city-db", "--geoip-db", dest="geoip_city_db", my_default='/usr/share/GeoIP/GeoLite2-City.mmdb',
>                          help="Absolute path to the GeoIP City database file.")
>         group.add_option("--geoip-country-db", dest="geoip_country_db", my_default='/usr/share/GeoIP/GeoLite2-Country.mmdb',
>                          help="Absolute path to the GeoIP Country database file.")
> ```
> 
> [Odoo 16.0 and before](https://github.com/odoo/odoo/blob/16.0/odoo/tools/config.py#L319)
> 
> ```
>          group.add_option("--geoip-db", dest="geoip_database", my_default='/usr/share/GeoIP/GeoLite2-City.mmdb',
>                          help="Absolute path to the GeoIP database file.")
> ```
> 
> As a workaround you should be able to add the following file to your odoo/custom/conf.d directory: `994_geoip.conf`
> 
> ```
> [options]
[...]

PR #601, opened 2024-03-04
[ADD] New geoip configuration file to provide odoo_version>=17.0 compatibility

josep-tecnativa:
CC @Tecnativa TT48174
In order to fix this issue, https://github.com/Tecnativa/doodba/issues/600 

Not tested yet.

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -106,6 +106,7 @@ RUN ln -s /usr/local/lib/python3.5/site-packages/doodbalib \
     /usr/local/lib/python3.5/site-packages/odoobaselib
 COPY build.d common/build.d
 COPY conf.d common/conf.d
+RUN rm common/conf.d/70-geoip.conf
 COPY entrypoint.d common/entrypoint.d
 RUN mkdir -p auto/addons auto/geoip custom/src/private \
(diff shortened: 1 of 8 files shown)


pedrobaeza:
This only applies to v17. Why are you doing this in all versions?

josep-tecnativa:
> This only applies to v17. Why are you doing this in all versions?
The changes I've made are to use the configuration file 60-geoip.conf for versions < 17.0, and for versions >= 17.0, to use 70-geoip.conf. To achieve this, for versions <17.0, I delete the file 70-geoip.conf:
`RUN rm common/conf.d/70-geoip.conf`
And for version 17.0, I delete 60-geoip.conf:
`RUN rm common/conf.d/60-geoip.conf`

ap-wtioit:
This would probably also need to be reflected in the [test code](https://github.com/Tecnativa/doodba/blob/master/tests/__init__.py#L612):
```python
                    # verify that geoip database is configured
                    (
                        "grep",
                        "-R",
                        "geoip_database = /opt/odoo/auto/geoip/GeoLite2-City.mmdb",
                        "/opt/odoo/auto/odoo.conf",
                    ),
```

pedrobaeza:
Superseded by #603

PR #604, opened 2024-03-20
[FIX] Fix docker image tags and ghcr.io auth credentials

PCatinean:
Hi @pedrobaeza small fix incoming:

1. The tags which are pushed should now be as before - {ODOO_VERSION}, {ODOO_VERSION}-onbuild and latest, latest-onbuild (where applicable)
2. The push to ghcr.io failed with the default credentials, I have switched them to the old ones hoping they will pass this time. If this also fails maybe you need to look into the permissions of the default github.repository_owner / secrets.GITHUB_TOKEN since authentication was sucessful.

Also after this works you would need to delete the extra {ODOO_VERSION}onbuild (with the missing -)

Fingers crossed :crossed_fingers:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -121,8 +121,8 @@ jobs:
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          username: ${{ secrets.BOT_LOGIN }}
+          password: ${{ secrets.BOT_TOKEN }}
       - name: Build and push
         uses: docker/build-push-action@v5
         with:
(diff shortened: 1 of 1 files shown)


PCatinean:
@pedrobaeza tests have passed, feel free to merge when you have some time

PCatinean:
@pedrobaeza it seems to have worked, all jobs passed and the images now have both architectures available for download. 

Last touch would be to delete the stale {ODOO-VERSION}onbuild images like these ones

Screenshot from 2024-03-20 19-07-59

pedrobaeza:
I'm not sure how the images are bridged to Docker Hub, as the workflow are supposed to be pushed to ghcr.io, and there, I don't see the other images: https://github.com/tecnativa/doodba/pkgs/container/doodba/versions?page=1

Do you know how this works?

PCatinean:
In the initial PR merged I had two bugs:

1. Added {ODOO_VERSION}onbuild instead of just {ODOO_VERSION}
2. Credentials used for ghcr.io were different and the user did not have permission to push

For this reason the {ODOO_VERSION}onbuild made it to the docker hub but not ghcr.io.

PR #608, opened 2024-05-02
remove -e from echo inside Dockerfile

### LINKED ISSUE #607: The generated /etc/adduser.conf may be corrupted
## Describe the bug

The generated /etc/adduser.conf has an additional `-e`, probably coming from the command `echo -e`
```
-e LAST_SYSTEM_UID=499              <----- here
LAST_SYSTEM_GID=499
FIRST_UID=500
FIRST_GID=500

```

The cause might be from here:
```
RUN echo -e "LAST_SYSTEM_UID=$LAST_SYSTEM_UID\nLAST_SYSTEM_GID=$LAST_SYSTEM_GID\nFIRST_UID=$FIRST_UID\nFIRST_GID=$FIRST_GID" >> /etc/adduser.conf \
```

## To Reproduce
no special steps, just build the image for version 17.0

**Affected versions**:

**Expected behavior** A clear and concise description of what you expected to happen.

**Additional context** Add any other context about the problem here. (e.g. OS, Docker
version, ...)

obayit:
This will still escape the \n to newline
closes #607

### DIFF
--- a/17.0.Dockerfile
+++ b/17.0.Dockerfile
@@ -40,7 +40,7 @@ ENV DB_FILTER=.* \
 
 # Other requirements and recommendations
 # See https://github.com/$ODOO_SOURCE/blob/$ODOO_VERSION/debian/control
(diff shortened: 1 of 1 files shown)


ap-wtioit (approved):
Introduced in https://github.com/Tecnativa/doodba/pull/587, unfortunately i only fixed 16.0.Dockerfile there. Thanks for the fix.

pedrobaeza (approved):

PR #609, opened 2024-05-03
Ignore manifest check for l10n_generic_coa in 17.0 and above

obayit:
Since Odoo 17.0 they removed the module `l10n_generic_coa`, but they kept its directory and with only the __init__.py file.

This causes manifest check to fail if `strict` was `True`.

This commit causes `addons_config` to ignore this module for 17.0 and above.

### DIFF
--- a/lib/doodbalib/__init__.py
+++ b/lib/doodbalib/__init__.py
@@ -129,7 +129,6 @@ def addons_config(filtered=True, strict=False):
                 skip_coa_17 = False
                 try:
                     odoo_version_float = float(ODOO_VERSION)
-                    skip_coa_17 = odoo_version_float >= 17 and addon_name == 'l10n_generic_coa'
                     skip_coa_17 = (
                         odoo_version_float >= 17 and addon_name == "l10n_generic_coa"
                     )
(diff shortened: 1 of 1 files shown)


obayit:
If this is merged, please combine the commits, to hide the "fix formatting" commits.

obayit:
To reproduce the "issue" run this command for a `17.0` deployment:
```
docker compose run --rm odoo addons init -x --with web
```

The `-x` will cause `strict` to be `True`.

pedrobaeza:
I think the leftover is more a bug than something intentional. I'm on travel for the Spanish OCA days, where I will meet with some people from Odoo SA. I'll check with them about this.

obayit:
Thanks @pedrobaeza , anyways I created a [PR](https://github.com/odoo/odoo/pull/164474) to remove that file from odoo/odoo

pedrobaeza:
For having the reference in both places: https://github.com/odoo/odoo/pull/164474

pedrobaeza:
Closing as the upstream has been merged. Thanks for the follow-up.

PR #622, opened 2024-09-26
Update 14.0.Dockerfile to build image after Postgres repository runs EOL

ananiasfilho:
Debian Buster Postgres repository was EOL (End Of Life). To make this image to continue working, was necessary pull Postgres packages from apt-archive repository.


zamberjo:
Same fix as in #623, I think the other PR is more complete.

pedrobaeza:
Let's go with the other if you don't mind, as CI is also being tested and here is red.

PR #627, opened 2024-10-07
[FIX] Use supported postgres version

josep-tecnativa:


pedrobaeza:
Why not finishing with the PG 16 support?

PR #628, opened 2024-10-08
[FIX] Not testing Doodba 18.0 to be able to push image

josep-tecnativa:
CC @Tecnativa TT51095

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,14 +35,11 @@ jobs:
       fail-fast: false
       matrix:
         # Test modern Odoo versions with latest Postgres version
-        odoo_version: ["18.0"]
-        pg_version: ["16"]
+        odoo_version: ["17.0"]
+        pg_version: ["15"]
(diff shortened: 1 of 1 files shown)


ap-wtioit:
@josep-tecnativa on our CI we first build images with a ci-nnnn- prefix and push them, then test them, and if the tests pass we tag them as the new images. would this workflow also work with github / dockerhub?

the other thing i discovered that might be related to the issues. in docker compose for about a year (when we switched from docker-compose to docker compose) it was no longer possible for us to use local only images that had a / in the name. so we needed to push the images to our registry first to be able to test them. testing a local copy of our.regsitry/path/to/doodba:nn.0-onbuild was no longer possible, it needed an image in the registry first.

josep-tecnativa:
I believe that workflow would indeed work with GitHub and Docker Hub. Once I finish updating to Doodba 18.0, I'll proceed with implementing it, as I see it as a very clean solution. Thank you!

PR #629, opened 2024-10-08
Add 18.0 (continuation)

josep-tecnativa:
CC @Tecnativa TT51095

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -106,6 +106,10 @@ jobs:
       DOCKER_REPO: tecnativa/doodba
       GHCR_HOST: ghcr.io
     steps:
+      - name: Clean Docker system before build
+        run: |
+          docker system prune -af || true
+          docker builder prune -af || true
       # Set build date as env variable to be used later
       - name: Set build date
         run: echo "BUILD_DATE=$(date '+%Y-%m-%dT%H:%M:%S.%N%:z')" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -35,23 +35,9 @@ jobs:
       fail-fast: false
       matrix:
         # Test modern Odoo versions with latest Postgres version
-        odoo_version: ["17.0"]
-        pg_version: ["15"]
+        odoo_version: ["18.0"]
+        pg_version: ["16"]
         python_version: ["3.10"]
-        include:
```
pedrobaeza, line 41: Why removing this?

tests/__init__.py
```diff
@@ -335,7 +335,7 @@ def test_addons_env(self):
                 ("test", "-e", "auto/addons/crm"),
                 ("test", "-d", "auto/addons/crm/migrations"),
             )
-        for sub_env in matrix(odoo_skip={"11.0", "12.0", "13.0", "17.0"}):
+        for sub_env in matrix(odoo_skip={"11.0", "12.0", "13.0", "18.0"}):
```
pedrobaeza, line 338: Why not testing all the possible versions (mostly recent ones - we can remove older ones)?

.github/workflows/ci.yaml
josep-tecnativa: This is for testing v18 only. It is a temporal commit

pedrobaeza (approved):

PR #647, opened 2025-03-07
feat(18.0): upgrade Python version to 3.12

zamberjo:
Update Python from 3.10 to 3.12 to match the official Odoo 18.0 Docker image

### DIFF
--- a/18.0.Dockerfile
+++ b/18.0.Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-slim-bookworm AS base
+FROM python:3.12-slim-bookworm AS base
 
 EXPOSE 8069 8072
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
AFAIK, 18.0 runbots (so the CI) runs Python 3.10, so changing this version is not convenient.

zamberjo:
Yes, I thought so, that's why I put the PR in draft. But I think it is a change worth considering.

pedrobaeza:
Not sure if the libraries versions are prepared for this Python version, and I also got a bad experience using the latest Python version due to a problem with the Odoo monkey-patches in gevent. That's why we always stick to the tested Python version.

PR #648, opened 2025-03-12
[IMP] Set gevent==21.12.0 & greenlet==1.1.0 for all architectures to ensure consistency

josep-tecnativa:
Fixed incompatibility error between gevent and greenlet:
`ValueError: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 152 from C header, got 40 from PyObject`
To ensure consistency and compatibility across architectures, we now set `gevent==21.12.0` and `greenlet==1.1.0` for both amd64 and arm64. This prevents binary mismatch issues and improves container stability.
CC @Tecnativa TT55390


pedrobaeza:
Does this only happens in 15.0?

josep-tecnativa:
I was just checking that, and it looks like it could also happen in 14.0. In Odoo > 15.0, gevent and greenlet versions are already controlled.

PR #653, opened 2025-04-28
Draft: Multiple Fixes from our Fork

ap-wtioit:
* base_search_fuzzy was migrated to 17.0 we can enable those tests now
* "18.0" is already skipped by `@prerelease_skip` so we do not need to add it to odoo_skip.
* ... (More to come if this test reveals that the fixes from our CI also are needed with github actions)
   * log_colorizer (wdb dependendency) issue with newest setuptools (2to3)
   * `--no-binary` test issue with `-r ...` requests dependency

Info @wt-io-it

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -29,7 +29,7 @@ GEIOP_CREDENTIALS_PROVIDED = environ.get("GEOIP_LICENSE_KEY", False) and environ
 # preparing the pre-release for the next version of Odoo, which hasn't been
 # released yet.
 prerelease_skip = unittest.skipIf(
-    ODOO_VERSIONS & {"16.0"}, "Tests not supported in pre-release"
+    ODOO_VERSIONS & {"18.0"}, "Tests not supported in pre-release"
 )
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Is this completed?

ap-wtioit:
The tests being successful in this PR was unexpected for me, i expected to see the same errors as on our CI but did not.

Enabling the tests for 18.0 is also done in #655, so my PR here is obsolete for now. I'll close this draft PR.

In the future (maybe in Belgium on the OCA days) i would like to propose to split up the test cases for doodba in smaller tests with a common base (e.g. the following pip.txt from dotd_ge_16 has unintended effects) and specific assertions for each test case. But that depends if you think that makes sense. dotd_ge_16 did fail to install dependencies on our CI but not on github after i prepared this PR.

The failing (for us) pip.txt:
```
# Compiled locally, which would fail if `apt_build.txt` fails
pycrypto==2.6.1 --no-binary :all:
# External dependency files can be included too; example: responses
-r https://raw.githubusercontent.com/OCA/server-auth/5f6272951d3fbd79ac39fde79c03832db46556b9/test-requirements.txt
```
From what i understood this `--no-binary :all:` is not only affecting pycrypto but is also used for test-reqiurements.txt which caused an error for us. So i think splitting up the 2 lines in 2 different test cases would make sense. And consequently we should also split up other tests (if we agree to split up the tests) so if they fail they should give a better indication of what is wrong.

pedrobaeza:
Aren't you coming to Spanish OCA days this year?

pedrobaeza:
Oh, what a pity! First year that you don't come since the debut of the Spanish OCA days...

PR #655, opened 2025-05-01
[UPD] Test environment

Tardo:
- Updated dependencies.
- All versions are now tested.
- Now “docker compose” (latest version) is used.
- The latest version of ubuntu is now being used.
- Resolve some 'TODO' comments

** As “docker-compose” is no longer installed (it is deprecated software), it is now required to have “docker compose” installed manually to launch the tests.


josep-tecnativa:
I added the PR Docker image push to enable easier end-to-end testing of proposed changes in PRs. In my opinion it doesn’t pose a security risk because we explicitly tag images as `tecnativa/doodba:<version>-pr-<PR number>-test`, making them clearly identifiable as test artifacts, so I believe this functionality shouldn’t be removed.

josep-tecnativa (approved):

pedrobaeza:
I don't agree with putting Ubuntu latest, as that will make red CI on next version. That version should be pinned until we explicitly change to next LTS, which is not the case. Please undo that change, @josep-tecnativa

PR #661, opened 2025-05-16
[FIX] Test Environment: Don't use published 'onbuild' image

Tardo:
Without these changes, the ‘onbuild’ docker hub image is used when launching the tests: https://hub.docker.com/r/tecnativa/doodba/tags

I consider this a bug. We are interested in testing the latest changes and testing the new changes on the new image, not on the last one that managed to be released. I see it as a mismatch between what I think I'm testing and what I'm actually testing.... and it can happen as now, that a problem is not detected when changes are made but when merging and launching new tests.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -70,14 +70,14 @@ files = [
 
 [[package]]
 name = "platformdirs"
-version = "4.3.7"
+version = "4.3.8"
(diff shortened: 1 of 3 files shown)


josep-tecnativa:
Why is CI green on this PR?

PR #665, opened 2025-06-16
[IMP] Add suppor for Pydevd/Pycharm debugging

david-banon-tecnativa:
Adds a wrapper to call pydevd-pycharm because it can't be called from cli like debugpy.
This pr goes alongside https://github.com/Tecnativa/doodba-copier-template/pull/542.

@tecnativa TT56757


josep-tecnativa (approved):

david-banon-tecnativa:
This will not be merged, we made the choice to use the newer debugpy DAP method added in 2026.1, as the issues i had have already been fixed, and doesn't require a pycharm pro subscription . A pr by one of my colleges will be open with the required changes in the near future.

PR #670, opened 2025-07-14
[FIX] Point Buster APT sources to archive

josep-tecnativa:
This PR rewrites the APT sources and disables Acquire::Check-Valid-Until, so 14.0 images build again.


13.0.Dockerfile
```diff
@@ -34,6 +34,8 @@ ENV DB_FILTER=.* \
     WDB_WEB_PORT=1984 \
     WDB_WEB_SERVER=localhost
 
+RUN sed -i 's,http://deb.debian.org,http://archive.debian.org,g;s,http://security.debian.org,http://archive.debian.org/debian-security,g;s,\(.*buster-updates\),#\1,;s,\(.*buster-backports\),#\1,' /etc/apt/sources.list && echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.conf.d/99no-check-valid-until
```
ap-wtioit, line 37: Do we need all of this? For us
```Dockerfile
# Debian buster was moved to archive
RUN sed -i 's,http://deb.debian.org,http://archive.debian.org,g;s,http://security.debian.org,http://archive.debian.org,g' /etc/apt/sources.list
```
like the statement from 12.0 seems to work fine (without the extra regexes for -updates, because buster-updates and buster-backports do exist in archive and do not need commenting out)

josep-tecnativa: You are right!

Thank you!

PR #674, opened 2025-09-01
[FIX][16.0][17.0][18.0] Dockerfile: restore single layer install

ap-wtioit:
with https://github.com/Tecnativa/doodba/pull/578 the images for 16.0 and 17.0 were changed to use 2 layers instead of one layer.

We could also keep the 2 layers starting with 16.0, but then i would suggest we check the cleanup and sync code at the end if it's not needed anymore or if it's needed twice.
```bash
    && apt-get autopurge -yqq \
    && rm -Rf wkhtmltox.deb /var/lib/apt/lists/* /tmp/* \
    && sync
```

Info @wt-io-it


josep-tecnativa:
This PR looks OK to me. I probably changed that by mistake at some point while debugging and accidentally pushed it.

PR #698, opened 2026-04-10
[IMP] Mount APT and PIP caches during build

david-banon-tecnativa:
### What:

This PR mounts APT and PIP caches during build time and a tmpfs at /tmp 

### Why:
This should speed up building images on local and during tests a little bit, as they will now only need to download (pip, apt) and build (pip) packages if they are a newer version.
This is done on v13+ only because older versions we don't test on CI shouldn't change if they work.

### Workflow changes
If you place things in /tmp during build and intend on using them later, this won't work anymore


yajo:
What happens if your host isn't debian-based?

david-banon-tecnativa:
> What happens if your host isn't debian-based?

It shouldn't matter. They content of cache mounts is only used during image build and is isolated from the host filesystem.

PR #700, opened 2026-05-27
[IMP] Add healthcheck

david-banon-tecnativa:
Adds a Healthcheck built into the image, this can be used in docker compose, kubernetes or other deployment systems to know if the container is actually working as expected.
@tecnativa @victoralmau @CristianoMafraJunior

### DIFF
--- a/bin/healthcheck
+++ b/bin/healthcheck
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+if (( ${ODOO_VERSION/%.0} > 13));then
+    curl -sf http://localhost:8069/web/health || exit 1
+else
+    curl -sf http://localhost:8069/web/database/selector || exit 1
+fi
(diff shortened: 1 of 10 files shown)


bin/healthcheck
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+if (( ${ODOO_VERSION/%.0} > 13));then
+    curl -sf http://localhost:8069/web/health || exit 1
```
ap-wtioit, line 4: this is a simple approach to a healthcheck for the odoo services that over time did not work for us.

there is cases where odoo http does not come alive for a long time and the service should still be marked as healthy (e.g. tests and db restores).

also there is cases were those 2 urls do not catch basic errors in odoo becoming unhealthy for endusers that the basic urls do not catch.

this is why i proposed #701, you can use it as it is or as inspiration.

Issue #702, opened 2026-07-06
Fixed: Debian Chromium 150.0.7871.46-1~deb12u1 is not working


ap-wtioit:
## Describe the bug

Odoo tests with chromium 150 (debian security release) are not working
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1141488

This is not really a doodba issue. But hopefully documenting it here helps someone not spending the time i just did.

Will close once debian has a new chromium package.

## To Reproduce

**Affected versions**: 16.0, 17.0, 18.0, 19.0

1. run doodba tests test_screencasts

**Expected behavior** A clear and concise description of what you expected to happen.

**Additional context** Add any other context about the problem here. (e.g. OS, Docker
version, ...)

pedrobaeza:
Ouch... any workaround for now?

ap-wtioit:
The currently built images should be fine to use for now. Our images from last week feature `Chromium 149.0.7827.196 built on Debian GNU/Linux 12 (bookworm)` the same as the official tecnativa doodba images (`docker run --rm --entrypoint="" tecnativa doodba:17.0 chromium --version`) . If debian doesn't fix it until tomorrow we could limit apt to an older version but `docker run --rm --entrypoint="" tecnativa/doodba:17.0 bash -c 'apt update && apt-cache madison chromium'` only shows
```
  chromium | 150.0.7871.46-1~deb12u1 | http://deb.debian.org/debian-security bookworm-security/main amd64 Packages
  chromium | 147.0.7727.137-1~deb12u1 | http://deb.debian.org/debian bookworm/main amd64 Packages
```
so we then might have to add additional repos to have 149.0.7827.196 still available to install. And the release from last week is missing security updates. 

That's why i prefer to wait a bit for a new release / official fix and not have an old chromium for too long.

ap-wtioit:
Other possible solutions would be:
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1141488#25 talks about ungoogled and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1141488#37 proposes that they send the patch upstream. I think this would be https://github.com/ungoogled-software/ungoogled-chromium/pull/3837/changes#diff-4487b42ff0bb51d1042b852f2a5af30200134050b8f97981154eced0a6ce03ceR11
```
[...]

PR #703, opened 2026-07-09
[IMP] tests: check chromium is working before testing artifacts with odoo

ap-wtioit:
A small improvement in test_screencasts checking chromium is actually working before testing odoo test-artifacts creation with base module install and odoo test_screencasts.

If chromium cannot show "about:blank" there is no need to test odoo screencasts.

Slight improvement when testing #702

Info @wt-io-it

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -857,6 +857,18 @@ class ScaffoldingCase(unittest.TestCase):
                     "{}",
                     ";",
                 ),
+                # check chromium version
+                ("chromium", "--version"),
+                # check that chromium is working before testing with odoo
+                (
+                    "chromium",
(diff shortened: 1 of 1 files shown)


tests/__init__.py
```diff
@@ -857,6 +857,21 @@ def test_screencasts(self):
                     "{}",
                     ";",
                 ),
+                # check chromium version
+                (
+                    "chromium",
+                    "--version",
+                ),
```
pedrobaeza: We can compact this removing trailing comma:

```suggestion
                ("chromium", "--version"),
```

pedrobaeza (approved):

david-banon-tecnativa (approved):

PR #710, opened 2026-08-14
[IMP] Use chromium on same milestone as OCA Google Chrome

david-banon-tecnativa:
This gets us more aligned with the OCA testing suite. See https://github.com/OCA/oca-ci/blob/master/Dockerfile#L97
This can be tested in projects by using the following image:
```
FROM ghcr.io/tecnativa/doodba:${ODOO_VERSION}-pr-710-test-onbuild
```

### DIFF
--- a/11.0.Dockerfile
+++ b/11.0.Dockerfile
@@ -6,6 +6,8 @@ ARG GEOIP_UPDATER_VERSION=4.1.5
 ARG MQT=https://github.com/OCA/maintainer-quality-tools.git
 ARG WKHTMLTOPDF_VERSION=0.12.5
 ARG WKHTMLTOPDF_CHECKSUM='1140b0ab02aa6e17346af2f14ed0de807376de475ba90e1db3975f112fbd20bb'
+# Milestone 126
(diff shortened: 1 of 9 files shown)


pedrobaeza (commented):
One question, can we pin the same versions, but with Chromium?

11.0.Dockerfile
```diff
@@ -6,6 +6,8 @@ ARG GEOIP_UPDATER_VERSION=4.1.5
 ARG MQT=https://github.com/OCA/maintainer-quality-tools.git
 ARG WKHTMLTOPDF_VERSION=0.12.5
 ARG WKHTMLTOPDF_CHECKSUM='1140b0ab02aa6e17346af2f14ed0de807376de475ba90e1db3975f112fbd20bb'
+# Mileston 126
```
pedrobaeza:
```suggestion
# Milestone 126
```

and the same for the rest

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

PR #725, opened 2026-09-02
[FIX] Upgrade curl version used in doodba

### LINKED ISSUE #724: Doodba builds fail fetching public GitHub repositories over HTTP/2 due to Debian 12 HTTP stack
## Describe the bug

Doodba currently uses Debian 12 in its base images. Since September 1, our Doodba builds started failing during 100-repos-aggregate when fetching public OCA repositories from GitHub.

We have isolated the problem to the HTTP/2 stack used by Git inside the Debian 12 based image.

## To Reproduce

The issue is reproducible outside git_aggregator on Debian 12:

```
GIT_TERMINAL_PROMPT=0 \
git -c http.version=HTTP/2 \
ls-remote https://github.com/OCA/contract.git HEAD
```
Result:

```
fatal: could not read Username for 'https://github.com': terminal prompts disabled
fatal: expected flush after ref listing
```
The repository is public and requires no authentication.

Forcing HTTP/1.1 works:

```
GIT_TERMINAL_PROMPT=0 \
git -c http.version=HTTP/1.1 \
ls-remote https://github.com/OCA/contract.git HEAD
d737f9711a3459e8f647ba48b146c4e5cd856e3f    HEAD
```
The affected Debian 12 environment uses:

git 2.39.5
curl/libcurl 7.88.1
nghttp2 1.52.0

We also compiled Git 2.48.1 on the same Debian 12 system and the problem remains, so upgrading Git alone does not fix it.

In comparison, on Debian 13:

git 2.47.3
curl/libcurl 8.14.1
nghttp2 1.64.0

the same Git HTTP/2 request works correctly.

Therefore this seems related to the HTTP/2/libcurl/nghttp2 stack inherited by the current Debian 12-based Doodba images.

This directly breaks Doodba builds because repository aggregation is performed during:

/opt/odoo/common/build.d/100-repos-aggregate

producing errors such as:

```
fatal: could not read Username for 'https://github.com': No such device or address
fatal: expected flush after ref listing

git_aggregator.repo contract
error calling ('git', 'fetch', 'contract', '...')
```
Would it be possible to update the affected HTTP libraries in the Doodba base image, move the base to a newer Debian release, or provide an official workaround that forces Git to HTTP/1.1 during repository aggregation?

**Affected versions**:

1. All Doodba images based on Debian 12

**Expected behavior** A clear and concise description of what you expected to happen.

The Docker image should build successfully

david-banon-tecnativa:
Upgrde from 7.88.1 to 8.14.1 to fix the following error:
`fatal: could not read Username for 'https://github.com': No such device or address`
Done on v16+ only atm as those are the easier ones to get working.
Fixes #724

### DIFF
--- a/16.0.Dockerfile
+++ b/16.0.Dockerfile
@@ -49,8 +49,10 @@ RUN --mount=target=/var/lib/apt/lists,type=cache,id=apt-lists-${TARGETARCH}-${OD
     && echo "SYS_UID_MAX $LAST_SYSTEM_UID\nSYS_GID_MAX $LAST_SYSTEM_GID" >> /etc/login.defs \
     && sed -i -E "s/^UID_MIN\s+[0-9]+.*/UID_MIN $FIRST_UID/;s/^GID_MIN\s+[0-9]+.*/GID_MIN $FIRST_GID/" /etc/login.defs \
     && useradd --system -u $LAST_SYSTEM_UID -s /usr/sbin/nologin -d / systemd-network \
+    && echo "deb http://deb.debian.org/debian bookworm-backports main" |  tee /etc/apt/sources.list.d/bookworm-backports.list \
     && apt-get -qq update \
-    && apt-get install -yqq --no-install-recommends curl \
+    && apt-get install -t bookworm-backports -yqq --no-install-recommends \
+        curl \
     && if [ "$TARGETARCH" = "arm64" ]; then \
         WKHTMLTOPDF_CHECKSUM=$WKHTMLTOPDF_ARM64_CHECKSUM; \
     elif [ "$TARGETARCH" = "amd64" ]; then \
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):

sergio-teruel (approved):

etobella:
@david-banon-tecnativa Even after using the new image, I still have the same problems... (v18, 19...)

I checked that the image used is the new one,

mariancuadranetkia:
@etobella we had same problem than you, but, now we checked again and works for us... I'm not sure the reason... but just now, works for us
@david-banon-tecnativa do you know the reason? As etobella, we had same problem one hour ago, but, now it's working.

ap-wtioit:
To also link the main github discussion here: https://github.com/orgs/community/discussions/206581

The Problem is not always happening but most of the times. So testing for solutions should be done carefully as it might seem what works at a certain time might not work 15 minutes later. And permanently adding fixes to images that only are needed temporarily may add additional maintenance cost for the images later.

etobella:
Ok, it seems fixed now... obviously this was something managed by github that was providing the problem.

I would keep monitoring and see what happens....

pedrobaeza:
Yes, I'm now starting to get the builds fixed, so indeed if this patch is not effective, we can revert it.

david-banon-tecnativa:
Yesterday i tried many times the following process and it worked, today not so much:

- On the debian bookwork and doodba base containers
- Clone a repo -> Asks for username
- Install bookwork backports and upgrade curl
- Clone a repo -> Clone works successfully

As the fix seems to be a fluke, i've tried the other route we know works, to configure git to use only http 1.1 in #728

pedrobaeza:
I'm afraid this is getting shape now to what I was fearing: they have limited the anonymous download:

https://github.com/OCA/helpdesk/pull/1100#issuecomment-5522360240

ValentinVinagre:
> I'm afraid this is getting shape now to what I was fearing: they have limited the anonymous download:
> 
> [OCA/helpdesk#1100 (comment)](https://github.com/OCA/helpdesk/pull/1100#issuecomment-5522360240)

😮‍💨 🤦🏻

PR #726, opened 2026-09-02
[FIX] Upgrade curl and git version of legacy odoo versions

david-banon-tecnativa:
Goal is similar to https://github.com/Tecnativa/doodba/pull/725, but this time for legacy versions, here we can't use backports as they are on non supported odoo versions, so we must build curl and git from source.


david-banon-tecnativa:
Closing as this wasn't the real fix to the original issue

PR #728, opened 2026-09-03
[IMP] Allow passing a GITHUB_TOKEN during downstream image builds

david-banon-tecnativa:
see https://github.com/Tecnativa/doodba/issues/724 and https://github.com/Tecnativa/doodba/issues/725

This works alongside https://github.com/Tecnativa/doodba-copier-template/pull/659/changes


pedrobaeza:
Wait, check https://github.com/OCA/helpdesk/pull/1100#issuecomment-5522360240

david-banon-tecnativa:
🫠
Fix an issue another one comes through

pedrobaeza:
Well, the final thing is that at the end, the problem is not curl, but it was only how it handles the error, redirecting to authenticate (although then it fails), or showing directly the error, but the underlying problem is that GitHub is limiting unauthenticated downloads, so although we do this or the other, we need to switch to authenticated downloads.

etobella:
@pedrobaeza If we want to switch to authenticated downloads I think that it is enough to modify the copier template. I made a test locally and worked

pedrobaeza:
But that only serves for local git-aggregate. For CI/CD systems, or servers, we would need more for not hardcoding credentials.

etobella:
I was talking about inserting the data into secrets :wink:

pedrobaeza:
Yes, if there's no other solution...

ap-wtioit:
github just confirmed that the http/1.1 workaround might be broken soon https://github.com/orgs/community/discussions/206581#discussioncomment-18264750:
> Setting http.version to HTTP/1.1 may currently work, but this could change without notice. We highly recommend using authentication, with or without updating this setting, to ensure continued access.

pedrobaeza:
They resist to confirm that there's now rate limit...

PR #701, opened 2026-05-27
[IMP] healthcheck: add healthcheck for odoo services

ap-wtioit:
@david-banon-tecnativa i saw your pull request #700 today. in our images we already have a more featured healthcheck in place. if you want you can use our version of the healthcheck as a starting point. it was improved over the years to feature a few more usecases for doodba. so if you are using healthchecks together with [autoheal](https://github.com/willfarrell/docker-autoheal) (or some equivalent) the failed healthchecks don't kill your long runnning click scripts / tests / db dump restores.

proposing our code seems the faster route than explaining all the edge cases in a review of #700.

Info @wt-io-it

PR #729, opened 2026-09-07
[FIX] 15.0: Pin security repo to pre-EOL snapshot

david-banon-tecnativa:
Pins bullseye security to a pre EOL snapshot while a proper solution is found.

TT64415

### DIFF
--- a/15.0.Dockerfile
+++ b/15.0.Dockerfile
@@ -32,6 +32,8 @@ ENV DB_FILTER=.* \
     WDB_WEB_PORT=1984 \
     WDB_WEB_SERVER=localhost
 
+# Debian bullseye is now EOL
+COPY apt_sources/bullseye.txt /etc/apt/sources.list
 # Other requirements and recommendations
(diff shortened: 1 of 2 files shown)


ap-wtioit:
Shouldn't we wait until bullseye security is moved to archive? Or is the plan to add them again later?

david-banon-tecnativa:
Odoo v15 isn't building properly as you migh have seen on https://github.com/Tecnativa/doodba/pull/701, this PR is a WIP fix in case we need to hotfix anything else. 
I don't have a lot of experience with the EOL process of debian, how long do they take to move EOL versions to archive?

ap-wtioit:
Well it is not that often that we need to change it. In the past the move happened smoothly we only needed to add the sed layer.

Currently https://archive.debian.org/debian-archive/debian/dists/ has bullseye. While security in archive doesn't: https://archive.debian.org/debian-archive/debian-security/dists/. Not sure if it is a bug or a process error.

What i fear is that by changing doodba 15.0 like this, the patch level is reduced from 31.8.2026 (EOL LTS) to 30.6.2024 (EOL) missing the security patches from about 2 years.

PR #732, opened 2026-09-18
[ADD] Doodba 20.0

david-banon-tecnativa:

PR #734, opened 2026-09-22
[IMP] 20.0: adapt tests and preparedb for new version

ap-wtioit:
Our fixes for the tests in our CI, without 20.0.Dockerfile that should be added by #732


david-banon-tecnativa:
Hi, i have added this commit onto #732, thanks for the work you put into this.

ap-wtioit:
Added in #732 with 24396d5640623ae5c4a3173971459cc5956958d7

